PR #114, opened 2016-05-19
Fixes to README examples and scripts

ijc25:
Various fixes and niggled which I tripped across while running through the README.

Possibly some of this is down to my using datakit in the basic mode rather than "experimental Github binding" mode?

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,9 +178,9 @@ contain:
 
 For example, to create a file `somefile`:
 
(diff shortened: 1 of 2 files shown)


samoht:
Thanks!

Indeed, the support for the Git http mode do indeed pull a lot more of dependencies (especially for TLS) so the default image does not include it. Only the `:github` image has it.

PR #132, opened 2016-05-24
Add a way to trigger remote fetches within the API

samoht:
Until now, triggering a `git fetch` requires the user the write the right runes in the filesystem. We now have the ability to trigger such fetches from within the API. This will be used by the Github bindings to fetch the PR branch under test.

### DIFF
--- a/src/ivfs/ivfs_remote.ml
+++ b/src/ivfs/ivfs_remote.ml
@@ -11,10 +11,28 @@ module Make (Store : Ivfs_tree.STORE) = struct
 
   module Sync = Irmin.Sync(Store)
 
-  type t = {
+  type remote = {
+    name: string;
+    url : string option;
+  }
+
+  type state = {
+    remote_name: string;
     remote_url : unit -> string option;
     update_head: Store.commit_id option -> unit;
+    fetch      : string -> unit Vfs.or_err;
   }
+
+  type t = {
+    mutable state: (state * Vfs.Inode.t) list;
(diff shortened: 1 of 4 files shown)


src/ivfs/ivfs_remote.ml
```diff
@@ -33,62 +51,94 @@ module Make (Store : Ivfs_tree.STORE) = struct
...
+      ) init;
+    t.root <- root;
+    t
+
+  let root t = t.root
+
+  let add t { name; url } =
+    t.state <- List.filter (fun s -> not @@ has_name name s) t.state;
+    let (_:Vfs.Inode.t) = inode_of_remote t ?url name in
+    ()
```
samoht: This doesn't match with the documentation string. We currently cannot easily change the remote url from an existing remote, so I'd probably change that code to follow what the doc says.

talex5:
BTW, in the client API I didn't provide access to named remotes at all, just a fetch function:

    val fetch : t -> url:string -> branch:string -> Commit.t or_error Lwt.t
    (** [fetch t ~url ~branch] fetches the given remote branch and returns its head commit. *)

https://github.com/docker/datakit/blob/1661a92f95b72053e1013632c2b134e2f1978667/src/client/datakit_S.mli#L191

The FS API, where the meaning of `head` depends on the last branch written to `fetch` seems hard to use safely (e.g. if a client has two threads, and wants to download two branches from the same remote at once).

samoht:
losing that PR for now on, until we have a better design.

PR #135, opened 2016-05-25
Minimal Go API and small tool (in Go) to synchronise GitHub API and Datakit

samoht:
This PR introduces a very simple Go API to use on top of Datakit. It is currently used in Docker for Mac and Windows Beta, so it is reasonnably well tested (although the unit-tests don't pass for `config.go` ...)

It also introduces a small tool to listen for GitHub webhooks and commit snapshots to Datakit. This is very/too much simple at the moment, but it demonstrates how to use the Go API to create transactions.

I am not a Go expert at all, so any feedback would be greatly appreciated. It would be especially nice to have a more idiomatic Go API (it that's possible).

### DIFF
--- a/api/go/client.go
+++ b/api/go/client.go
@@ -1,0 +1,265 @@
+package datakit
+
+import (
+	"io"
+	"log"
+	"net"
+	"sync"
+
+	p9p "github.com/docker/go-p9p"
+	"golang.org/x/net/context"
+)
+
+type Client struct {
+	conn     net.Conn
+	session  p9p.Session
+	m        *sync.Mutex
+	c        *sync.Cond
+	usedFids map[p9p.Fid]bool
+	freeFids []p9p.Fid
+	root     p9p.Fid
+}
+
+var badFid = p9p.Fid(0)
+
+var rwx = p9p.DMREAD | p9p.DMWRITE | p9p.DMEXEC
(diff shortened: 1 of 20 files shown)


api/go/client.go
```diff
@@ -0,0 +1,268 @@
...
+	"log"
+	"net"
+	"sync"
+
+	p9p "github.com/docker/go-p9p"
+	"golang.org/x/net/context"
+)
+
+// TODO: replace this with a variant of the new filesystem interface
+// see issue #90
```
talex5: Issue #90 is something else...

samoht:
Build error is the for `github` bindings due to Inria servers outage.

src/bin/main.ml
```diff
@@ -233,7 +241,45 @@ let start urls sandbox git ~bare =
...
+            )
+        | Some path ->
+          Lazy.force Git_fs_store.listener;
+          let prefix = if sandbox then "." else "" in
+          let path = prefix ^ path in
+          let push () =
+            Logs.debug (fun l -> l "Pushing %s to %s" path remote);
+            let cmd =
+              Lwt_process.shell @@
+              Printf.sprintf "cd %s && git push %s --all" path remote
```
talex5: The `%s` here looks risky. At least `%S` would work with spaces. Annoying that `Lwt_process` doesn't provide a portable way to set the directory.

PR #136, opened 2016-05-25
Use latest ASL library to avoid runtime errors on startup

samoht:


samoht:
Merging this PR as the build errors are due to Inria server outage (http://gallium.inria.fr/~fpottier/ is not available) and concern only the GitHub bindings.

PR #143, opened 2016-06-01
Improve speed of appending to large files

### LINKED ISSUE #102: Make appending efficient
The Linux 9p client writes large files (e.g. the Moby iso) 8KB at a time. Datakit handles each write by allocating a new buffer, copying the old version into it, and adding the new 8KB. This gets slow quickly ;-)

Change the type for files in `Ivfs_tree` from

      | Blob of Cstruct.t

to

      | Blob of Blob.t

and make a suitable `Blob` module for efficient non-destructive updates. e.g. copy-on-write, or just use a list of blocks in reverse order (and concat them all together at the end).

talex5:
Replace Cstruct.t with Ivfs_blob.t, which allows fast appending.

In a simple benchmark, writing a 102400 KB file in 800 KB chunks went from taking 436 to 2.8 seconds.

Fixes #102.

### DIFF
--- a/src/ivfs/ivfs_blob.ml
+++ b/src/ivfs/ivfs_blob.ml
@@ -1,0 +1,85 @@
+open Result
+
+type t = Cstruct.t list ref (* (reversed) *)
+
+let ( >>!= ) x f =
+  match x with
+  | Ok x -> f x
+  | Error _ as e -> e
+
+let len t = Cstruct.lenv !t
+
+let empty_cs = Cstruct.create 0
+let empty = ref []
+
+let of_string s = ref [Cstruct.of_string s]
+
+let of_ro_cstruct cs = ref [cs]
+
+let to_ro_cstruct t =
+  let cs = Cstruct.concat (List.rev !t) in
+  t := [cs];
+  cs
+
+let to_string t =
+  Cstruct.to_string (to_ro_cstruct t)
+
+(* [overwrite orig (new, offset)] is a buffer [start; padding; new;
+    end] where [new] is at position [offset], [start] and [end] are
+    from [orig] and [padding] is zeroes inserted as needed. *)
+let overwrite orig (data, offset) =
+  let orig_len = Cstruct.len orig in
+  let data_len = Cstruct.len data in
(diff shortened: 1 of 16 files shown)


avsm:
Nice speedup!

src/ivfs/ivfs_blob.ml
```diff
@@ -0,0 +1,83 @@
...
+  | 0L -> Ok empty
+  | new_len ->
+    let new_len = Int64.to_int new_len in
+    let extra = new_len - len old in
+    if extra = 0 then Ok old
+    else if extra < 0 then (
+      Ok (of_ro_cstruct (Cstruct.sub (to_ro_cstruct old) 0 new_len))
+    ) else (
+      let padding = Cstruct.create extra in
+      Cstruct.memset padding 0; (* Won't be needed with Cstruct >= 2.2 *)
```
avsm: I can release cstruct 2.2 now and then we can remove this and make that a hard constraint... is anyone else blocking cstruct 2.2?

src/ivfs/ivfs_blob.mli
```diff
@@ -0,0 +1,42 @@
...
+
+val of_ro_cstruct: Cstruct.t -> t
+(** [of_ro_cstruct c] is a blob containing the data [c].
+    Note that we take a reference to [c] rather than copying, so [c] MUST NOT be modified
+    after this call. [c] may also be shared with modified versions of the resulting blob. *)
+
+val to_ro_cstruct: t -> Cstruct.t
+(** [to_ro_cstruct t] is a read-only Cstruct with the same data as [t].
+    DO NOT modify this - it may corrupt the blob or other blobs sharing data
+    with this one if you do. *)
```
avsm: Worth an feature request issue on Cstruct for properly typed RO buffers in a future revision?

talex5: Would be nice. I've added it to the long-standing phantom-type issue.

samoht:
LGTM

avsm:
LGTM

PR #93, opened 2016-05-10
Add initial draft for dkt command-line tool

samoht:
Initial draft to start the discussion

PR #146, opened 2016-06-08
Initial design draft for dkt command-line tool

samoht:
Subsume #93 

Note: there are still unresolved comments on the previous PR, I plan to address them in a more substantial re-design of the tool.

### DIFF
--- a/docs/dkt.md
+++ b/docs/dkt.md
@@ -33,7 +33,7 @@ while keeping full provenance and reproducibility.
 To create a new datakit project, run:
 
 ```
-$ dkt init [--github-token=TOKEN] [--docker-token=TOKEN]
+$ dkt init [--github-token=TOKEN] [--hockeyapp-token=TOKEN]
 Initialized empty Datakit repository in /tmp/foo/.dkt/
 ```
 
@@ -53,7 +53,8 @@ $ dkt clone REPO [--github-token=TOKEN] [--docker-token=TOKEN]
 To add a new source:
 
 ```
-dkt source add NAME LOCATION[:TAG] [-k KIND] [-p PATH]
+$ dkt source add NAME [LOCATION[:TAG]] [-k KIND] [-p PATH]
+    [--read-only] [--write-only] [--read-write]
 ```
 
 Usually the kind can be infered from the location, and the location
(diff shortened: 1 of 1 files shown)


docs/dkt.md
```diff
@@ -0,0 +1,341 @@
...
+`--read` and `--write`
+
+As an example, let start by describing some of the data sources needed
+to build and test [pinata](https://github.com/docker/pinata) today:
+
+```
+$ dkt source add pinata docker/pinata --king github
+$ dkt source add pitfall docker/pinata:master --path v1/cmd/pitfall --read-only
+$ dkt source add cases docker/pinata --path v1/tests
+$ dkt source add dmg e63b4e58fd9d40e48e74f8f2e71c49a5 --kind hockeyapp
```
rneugeba: shouldn't this be `docker/pinata:master --path /tests`?

docs/dkt.md
```diff
@@ -0,0 +1,341 @@
...
+vary depending on the source kind. For instance, for Git it is
+`--read` and `--write`
+
+As an example, let start by describing some of the data sources needed
+to build and test [pinata](https://github.com/docker/pinata) today:
+
+```
+$ dkt source add pinata docker/pinata --king github
+$ dkt source add pitfall docker/pinata:master --path v1/cmd/pitfall --read-only
+$ dkt source add cases docker/pinata --path v1/tests
```
rneugeba: shouldn't this be docker/pinata:master --path /tests?

docs/dkt.md
```diff
@@ -0,0 +1,341 @@
...
+state. Compute node can have OS/architecture constraints, and can be
+successful or fail.
+
+To use the `pitfall` tool, we create two phases: one for build and one
+for run, and we create a temporary local path which will contain the
+build artefacts generated during the build phase:
+
+```
+$ dkt source add build
+$ cat buid.yml
```
avsm: s/buid.yml/build.yml

talex5:
Are we still planning to do this, or does the CI stuff cover what we need now?

samoht:
I still think we need a CLI tool to interact with the CI (that we don't have yet).

Also I think there are still few things we are missing. The main one being: how does the current CI expose the data generated in one CI step to the next one? (e.g. if we want to analyse the logs for instance?

samoht:
But yes your CI library covers already a large bunch of that. 

An other point: would be nice to have a simple file format to express simple CI pipelines.

talex5:
It's just a monad, so you can move data around as with any program. e.g.

    step1 >>= fun output_of_step1 ->
    step2 output_of_step1

(or just `step1 >>= step2` of course)

There's currently no API call to get the logs from the previous step, but no reason why we couldn't provide one (although you'd actually get the tree of logs for all steps leading to this one). Or maybe filtering should be done by the plugin (e.g. `run ~filter_logs cmd`) so you process the actual output from the command, without any extra logging the CI does.

samoht:
Would it be easy/useful for instance to have high level constructs to start a new container with the data from the last one bind-mounted at a given path (so it can be accessible by the container easily)?

talex5:
Yes. At least, the result of a `docker build` step should be the image ID, which you can then use in the next `docker build` or `docker run`, for example. Could do something similar with Git repositories, where each step produces a commit ID. And, for smaller results, we could store them directly in DataKit.

But in any case, this PR doesn't seem very related to extending the CI, so perhaps it should be closed and the improvements mentioned in this thread turned into feature request issues?

samoht:
Closing as that doc is now quite outdated.

PR #148, opened 2016-06-09
Add OCaml client bindings

talex5:
This PR adds an OCaml client library to DataKit.

The API is described in https://github.com/talex5/datakit/blob/ocaml-client/api/ocaml/lib/datakit_S.mli
An example is in https://github.com/talex5/datakit/blob/ocaml-client/examples/ocaml-client/example.ml

The API shares some similarities with the CueKeeper Irmin wrapper (https://github.com/talex5/cuekeeper/blob/master/utils/git_storage_s.ml). It is also based in part on the `test_utils.ml` code. I have ported the tests to the new API (as `test_client.ml`), but kept the originals too as they test some misuse of the 9p API that the client bindings prevent.

All features of the main API should be available, although some shell conveniences aren't present (e.g. in the shell you can `cd /branch/foo/ro` and get a read-only-but-changing view of the branch. In the API you must `DK.Branch.head foo` and then work on that one commit. This should help prevent races.

I also implemented the `/remotes` part too, but I don't think that is fully working yet. The github integration is not implemented yet.

Hopefully these bindings provide a better example for people writing bindings for other languages (the previous Go bindings were hastily copied from Docker4Mac and are rather specific to that use).

/cc @dsheets (who was asking about this), @dave-tucker (who is working on C# bindings) and @djs55 (who wrote the Go bindings).


samoht:
> Move OCaml client under src so docs build

What was the error message? Would be great if all the client bindings could live at the same place (maybe add a symlink if that's not possible?)

talex5:
The error was:

> Solver failed:
> Ocamlbuild knows of no rules that apply to a target named ../api/ocaml/lib/Datakit_client_9p.mly. This can happen if you ask Ocamlbuild to build a target with the wrong extension (e.g. .opt instead of .native) or if the source files live in directories that have not been specified as include directories.

samoht:
LGTM

PR #150, opened 2016-06-15
Handle malformed listen addresses better

talex5:
Before, if you specified "tcp:192.168.1.2" it would actually listen on
127.0.0.1 (because the URL is malformed). We now reject this.


samoht:
LGTM

PR #158, opened 2016-06-23
Add missing 'opam update' to Alpine fix

talex5:
OK, I tested with `--no-cache` this time...


talex5:
(merging now because CI doesn't test the Docker scripts anyway)

PR #161, opened 2016-06-27
Switch to topkg

samoht:


_tags
```diff
@@ -0,0 +1,40 @@
+true : bin_annot, safe_string, package(bytes)
```
avsm: could you add `debug` here as well?

avsm: and move all the `package()` declarations to the same part of the tags file?

src/irmin-io/ir_io.mli
```diff
@@ -0,0 +1,7 @@
+module Sync: Git.Sync.IO
```
avsm: missing license?

_tags
dbuenzli: Regarding `debug` feed back about https://github.com/dbuenzli/topkg/issues/54 is welcome.

src/irmin-io/ir_io.mli
samoht: The apache license says that you should only add a license file at the root of the repo. I'm happy to add license headers in the file too if that's necessary.

PR #163, opened 2016-06-29
Fix Makefile and Dockerfiles

samoht:


Dockerfile
```diff
@@ -1,22 +1,14 @@
...
-RUN cd /home/opam/opam-repository && git reset --hard dd38f0690ff01c8e1ed0eed231307ba4d00c50df && opam update
-RUN opam depext lwt ssl &&  opam install lwt alcotest oasis
-RUN opam pin add asl.dev https://github.com/mirage/ocaml-asl.git
-#RUN cd /home/opam/opam-repository && git pull && opam update
+RUN sudo apk add ncurses-dev
+RUN opam depext lwt && opam install lwt alcotest
 
 RUN opam pin add hvsock https://github.com/djs55/ocaml-hvsock.git
 
-COPY opam /home/opam/src/datakit/opam
```
talex5: I guess the idea here was that you don't have to reinstall all the dependencies except when the `opam` file changes...

Dockerfile.client
```diff
@@ -1,19 +1,14 @@
...
-COPY src /home/opam/src/datakit/src/
-COPY .git /home/opam/src/datakit/.git
-COPY Makefile /home/opam/src/datakit/Makefile
+COPY . /home/opam/src/datakit
 
 RUN sudo chown -R opam.nogroup /home/opam/src
 WORKDIR /home/opam/src/datakit
 
-RUN make src/bin/version.ml
+RUN rm _tags
```
talex5: What's this for?

Dockerfile
samoht: I am not sure we want to use the Dockerfile in dev mode. I think keeping these files simple is better.

Dockerfile.client
samoht: otherwise it will try to pick the `true: -package(..)` define in the `_tags` files.

PR #165, opened 2016-06-30
Use upstream hvsock instead of the dev version

samoht:


samoht:
This should add more green ticks in https://hub.docker.com/r/docker/datakit/builds/

PR #167, opened 2016-07-01
Read GitHub persistent events to populate a Git branch

samoht:
See the commit description and the README updates.

@talex5 is that ok for you? You should be able to write `github-hooks` in the `/sync` file to merge persistent and webhooks events in the same branch. Hopefully I picked the same paths for both kind of events ...


samoht:
Note: that implementation is horrible hack, as a subtree of the VFS layer (e.g. `/db/github.com`) reconnects to the root of DataKit using a 9p client. The proper way to do this is to lift the client API at the VFS level.

talex5:
Why is this in DataKit itself? Shouldn't all this be in `gh-hooks`?

samoht:
ideally yes... But how do you tell gh-hooks to reload (e.g. read `/github.com/user/repo/events`) and/or how does gh-hook knows which `user/repo` to look at? The hooks are currently manually set outside of gh-hooks.

src/vgithub/vgithub.ml
```diff
@@ -317,42 +272,139 @@ let pr_dir t pr =
...
+          | `Closed ->
+            DK.Transaction.exists tr dir >>*= fun exists ->
+            if exists then DK.Transaction.remove tr dir else ok
+          | `Open   ->
+            DK.Transaction.make_dirs tr dir >>*= fun () ->
+            let data = Cstruct.of_string pr.pull_head.branch_sha in
+            DK.Transaction.create_or_replace_file tr ~dir "head" data
+        in
+        let update_status tr s =
+          let dir = root / "commits" / s.status_event_sha in
```
samoht: I guess this path is wrong and should contain the status name too.

samoht:
(also I am a bit worried by doing too much high-level stuff in gh-hooks today as the Go API is not at the same level as the OCaml one yet)

talex5:
`gh-hooks` already seems to know what to monitor (though I'm not sure how). I'd assume it would have a config file in the `github-hooks` branch.

I was expecting that I wouldn't need to tell it to reload - making sure the data is up-to-date is the whole point of its existence...

samoht:
no, gh-hooks just receives webhooks event and process them without checking anything special (as long as the secret is correct).

You can trigger a `sync` when `pinata-ci` starts and have a button in the "UI" to use when there are obvious missing events (because `gh-hooks` missed some events for instance?)

samoht:
Or maybe `gh-hooks` can trigger the sync itself on startup?

talex5:
pinata-ci starting doesn't have anything to do with whether the information from gh-hooks is up-to-date, so I don't think that helps much.

It would make more sense to trigger the update when gh-hooks starts, since that's when we know we might have lost events.

In general, I think pinata-ci should only ever talk to the database. All interaction with github should be the responsibility of some other component. If it's difficult to write the sync code in Go, we could put an OCaml client in the gh-hooks container, which the Go code could run directly perhaps?

samoht:
Let's close this, I'll open a better PR with the correct architecture soon.

PR #170, opened 2016-07-04
Fix races in stream handling

talex5:
- Reading on a stream returned the next value after the read,
  not the next value since the previous read.

- Added unit-test.

/cc @samoht


samoht:
woops... good catch!

PR #172, opened 2016-07-05
Make sure libev is available

### LINKED ISSUE #152: Unix.Unix_error(Unix.EINVAL, "select", "") on Alpine Linux
Looks like lwt doesn't detect `libev` on Alpine 3.3 and falls back to select mode. Perhaps select mode is buggy (I guess not many people use it)?

talex5:
Fixes #152.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM ocaml/opam:alpine
 
-RUN sudo apk add ncurses-dev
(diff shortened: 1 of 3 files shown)


talex5:
This performs the test @djs55 suggested in the Dockerfile to make sure we don't build a binary that uses `select`. This fix shouldn't be needed once @avsm has updated the `ocaml/opam:alpine` image to support libev by default, but doesn't do any harm.

PR #174, opened 2016-07-05
Datakit / Github API bridge

samoht:
This PR allows to interact with the GitHub API using only a Git branch and Git operations. This is still experimental and needs more tests and fixes, but the main part of the feature is there.


samoht:
I start to be happy with the unit tests: repositories are monitored properly and user changes are reflected back as GitHub API calls (only the status updates currently). Next step is ignoring the changes induced by the webhook (to avoid triggering unnecessary API calls) and test end-to-end deployments. Will do that tomorrow. Given the complexity of the beast, I'd say this might be ready on Friday.

PR #175, opened 2016-07-06
Cleanups

samoht:


samoht:
@talex5 875bdca introduces a breaking API change, you might need to upgrade your libraries.

PR #177, opened 2016-07-11
[do not merge] Use the released-but-not-available hvsock.0.8.1

djs55:
This will trigger appveyor to create an artifact which we can test,
while waiting for the hvsock.0.8.1 package metadata to slowly percolate
around the various package universes.

### DIFF
--- a/_tags
+++ b/_tags
@@ -22,7 +22,7 @@ true: \
 
 <src/conduit/*>: \
   package(threads), thread, \
-  package(conduit.lwt-unix), package(hvsock.lwt), \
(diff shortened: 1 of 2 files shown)


samoht:
@djs55 can you rebase that PR now that everything is in opam?

samoht:
@djs55 do we want to integrate the latest fixes in hvsock in datakit too?

djs55:
I don't think it's critical for data kit (single long-lived connection which doesn't use the special `Flow_lwt_hvsock_shutdown` module) but it would be nice to have everyone on the same version.

I've just made [ocaml/opam-repository#7051]

samoht:
Thanks! I have opened #204 to follow-up with your changes.

PR #181, opened 2016-07-14
Don't log at info level every time we read or write a file

talex5:


avsm:
lgtm!

PR #182, opened 2016-07-14
Use trusty for testing

samoht:


talex5:
And the examples...?

talex5:
Seems there's something odd about io-page in general. e.g. https://github.com/mirage/irmin/issues/338
I guess we have an external used in io-page that's only provided by io-page.unix :-(

djs55:
That's "the linking hack" for you. Perhaps there's another way to slice it -- IIRC the stub on Unix calls `posix_memalign` (and something else on Xen). Perhaps we should define `posix_memalign` in minios so we can use one stub everywhere?

samoht:
My take on this is https://github.com/mirage/ocaml-9p/pull/91

samoht:
I've reverted most of the commits from that PR and published on new version of `ocaml-9p` which should fixes the initial issue upstream. I've also renamed the PR's title to reflect the new reality of that PR.

talex5:
Seems the new protocol-9p is now missing a dependency on `cstruct.lwt`...

talex5:
Fix for missing cstruct.lwt: https://github.com/mirage/ocaml-9p/pull/93

samoht:
I don't understand why I can't repro these errors on my local machines :-( I've made a new [release](https://github.com/ocaml/opam-repository/pull/6980) of ocaml-9p, let's hope it works this time.

talex5:
I can repro it on my OS X laptop.

samoht:
Fixed on Linux/OSX, on windows, we need procotol-9p 0.7.2 to reach @fdopen's repository.

PR #184, opened 2016-07-18
Gh bridge

samoht:


src/bin/github_bridge.ml
```diff
@@ -65,42 +76,57 @@ let start () sandbox listen_urls
...
   let connect_to_datakit () =
-    Log.info (fun l -> l "Connecting to %s" datakit);
     let proto, address = parse_address datakit in
-    Client9p.connect proto address () >>= function
+    Log.info (fun l -> l "Connecting to %s (%s, %s)" datakit proto address);
+    (Lwt.catch
+       (fun () -> Client9p.connect proto address ())
+       (fun _ -> Lwt.fail_with
+           "%s is not a valid connect adress. Use 'tcp:hostname:port' or \
+            'unxi:path'."))
```
talex5: Typo: `unxi`

src/client/datakit_client_9p.ml
```diff
@@ -577,6 +577,8 @@ module Make(P9p : Protocol_9p_client.S) = struct
         (fun hash -> node_of_hash t hash >>*= fn)
 
     let fast_forward t commit =
+      let path = Datakit_path.of_steps_exn (branch_dir t) in
+      FS.make_dirs t.fs ~base:[] path >>*= fun () ->
```
talex5: Why is this needed? We already ensure the directory exists in `create`.

samoht: It was failing if the branch didn't exist. Are you sure we are doing the right test in `create`?

talex5: `test` doesn't do a create, it just creates the directory:

```
    let create fs name =
      (* Note: DataKit returns success if the branch already exists too,
         so no need to handle errors here. *)
      FS.create_dir fs ~dir:["branch"] name >|*= fun () ->
      { fs; name }
```

The only way I can see it failing is if someone deleted the branch after the create. Would be interesting to see a trace if you can reproduce it.

samoht:
I think this is ready to merge! I will try to deploy that tomorrow morning.

PR #186, opened 2016-07-18
Use CircleCI to build and store OSX artifacts

samoht:


Makefile
```diff
@@ -2,11 +2,17 @@ GITHUB=$(shell opam config var github:installed)
 APP=Datakit.app
 EXE=Datakit
 
+COMMIT_ID=$(shell git rev-parse HEAD)
```
talex5: Is this going to break builds from release archives?

samoht: The release archives should not call `make COMMIT` (hopefully). As `make` is lazy, that should be fine.

ijc25: To be double+ sure you can do:
```
COMMIT: COMMIT_ID=$(shell....)
COMMIT:
       @echo $(COMMIT_ID) > COMMIT
```
so the var is only defined for that target, but I'm not sure why it isn't just:
```
COMMIT:
        @git rev-parse HEAD > COMMIT
```
in the first place.

PR #188, opened 2016-07-19
Fix default datakit address in Dockerfile.github

samoht:
This is pretty confusing that the listen and connect patterns are different.

/cc @djs55 and @dsheets


dsheets:
Where is the parser code?

samoht:
I've copied the code from ocaml-9p, but basically here is [connect](https://github.com/docker/datakit/blob/master/src/bin/github_bridge.ml#L42) and there is [accept](https://github.com/docker/datakit/blob/master/src/conduit/datakit_conduit.ml#L149)

samoht:
btw, this is also different from the Go `dial` patterns ...

PR #192, opened 2016-07-19
Use maximum_write_payload funtion in recent ocaml-9p

talex5:
Allows much larger writes (was limited to 8K per message).


talex5:
CI failed due to depext waiting for input. I assume this is the same issue as https://github.com/mirage/mirage-www/issues/474#issuecomment-233621222, which @avsm is working on in https://github.com/ocaml/opam-depext/pull/52.

avsm:
See ocaml/opam-repository#7016

PR #193, opened 2016-07-20
Improve the github bridge

samoht:


src/vgithub/vgithub.ml
```diff
@@ -781,30 +788,36 @@ module Sync (API: API) (DK: Datakit_S.CLIENT) = struct
...
       DK.Branch.with_transaction pub (fun tr ->
           DK.Transaction.merge tr c >>*= fun (_, conflicts) ->
           if conflicts <> [] then failwith "TODO";
           let msg = Fmt.strf "Merging with %s" @@ DK.Branch.name priv in
-          DK.Transaction.commit tr ~message:msg
-        ) >>*= fun _ ->
-      ok ()
+          of_commit c >>*= fun priv ->
+          of_tree (E ((module DK.Transaction), tr)) >>*= fun pub ->
+          (if compare_snapshot old_t.priv priv = 0 then
```
samoht: @talex5 I think this does the right thing: e.g. it merges only if the private branch has changed.

talex5: Looks right. I'll add an issue to support this directly in DataKit though.

hooks/datakit-gh-hooks/server/event.go
```diff
@@ -112,7 +117,7 @@ func (h *Server) CommitDir(e github.StatusEvent) ([]string, error) {
 	sha := *e.SHA
 
 	h.logger.Debugf("user=%s, repo=%s", user, repo)
-	commitDir := []string{user, repo, "commits", sha}
+	commitDir := []string{user, repo, "commit", "status", sha}
```
talex5: Should this be named `statusDir` then?

tests/test_github.ml
```diff
@@ -240,6 +239,11 @@ let test_events dk =
   Alcotest.(check counter) "counter: 0"
     { Counter.events = 0; status = 0; set_pr = 0; set_status = 0 } t.API.ctx;
   VG.sync ~policy:`Once s ~priv ~pub ~token:t >>= fun s ->
+  VG.sync ~policy:`Once s ~priv ~pub ~token:t >>= fun s ->
+  VG.sync ~policy:`Once s ~priv ~pub ~token:t >>= fun s ->
+  VG.sync ~policy:`Once s ~priv ~pub ~token:t >>= fun s ->
+  VG.sync ~policy:`Once s ~priv ~pub ~token:t >>= fun s ->
+  VG.sync ~policy:`Once s ~priv ~pub ~token:t >>= fun s ->
```
talex5: Is this intentional or a copy-and-paste error?

samoht: it's intentional, I want to really be sure that no actions is taken :-)

samoht:
I've renamed the variable to `statusDir` and rebased to master. I have also added 2 more commits:
- avoid modifying the whole world on startup for it to look like `github-metadata`. It's just confusing, and even more if it's revert back some changes made by the users. It also seems to change the owner of every "status" on the open PRs, which can be confusing as well.
- add a `--dry-updates` command-line argument to the GH bridge, to run it in "read-only" mode.

samoht:
There is still one missing commit that I would like to make:

- would be great to create only one transaction in the private branch too instead of make each phases create in a separate commit, while at the same time the GH webhook continue to write in the branch (so possible races there too).

Anyway, this is not a blocker, so I'll merge the PR, try to deploy it and make another PR with that fix later.

src/vgithub/vgithub.mli
```diff
@@ -118,12 +118,14 @@ module Sync (API: API) (DK: Datakit_S.CLIENT): sig
   (** Create an empty sync state. *)
 
   val sync:
-    ?switch:Lwt_switch.t -> ?policy:[`Once|`Repeat] ->
+    ?switch:Lwt_switch.t -> ?policy:[`Once|`Repeat] -> ?dry_updates:bool ->
     pub:DK.Branch.t -> priv:DK.Branch.t -> token:API.token ->
     t -> t Lwt.t
 (** [sync t ~pub ~priv ~token] mirror GitHub changes in the DataKit
     public branch [pub]. It uses the private branch [priv] to store
     the received wehook event states. It connects to the GitHub API
```
talex5: Typo: `wehook`

src/vgithub/vgithub.ml
```diff
@@ -549,82 +547,121 @@ module Sync (API: API) (DK: Datakit_S.CLIENT) = struct
...
-    user_repos: UserRepoSet.t;            (* active hooks, computed from tree *)
-    hooks     : HookSet.t;         (* user_repo + FS tree, computed from tree *)
-  }
+  let empty_snapshot =
+    { repos = XRepoSet.empty; status = XStatusSet.empty; prs = XPRSet.empty }
 
-  let empty =
-    { hooks = HookSet.empty; user_repos = UserRepoSet.empty;
-      tree = None; commit = None }
+  type t = { pub: snapshot; priv: snapshot; synced: bool }
```
talex5: Would be good to have a comment here saying what `synced` means.

src/vgithub/vgithub.ml
```diff
@@ -770,41 +785,49 @@ module Sync (API: API) (DK: Datakit_S.CLIENT) = struct
...
+      Log.debug (fun l -> l "user-repo-diff: %a" XRepoSet.pp diff);
       ok diff
     in
-    let github_calls last_t current =
-      of_commit current >>*= fun current_t ->
-      call_github_api ~token ~old:last_t current_t >>*= fun () ->
-      ok ()
-    in
+    let call_github_api old_t current =
+      if not t.synced then ok ()
```
talex5: Assuming `synced` means GitHub is already up-to-date, this test seems to be the wrong way around.

samoht:
@talex5 I've renamed `synced` into `merged` and try to add some comments. Is that clearer?

talex5:
Yes, thanks!

PR #196, opened 2016-07-20
Update Go web-hooks to use new layout

talex5:
Untested. @samoht: does this look right to you?


talex5:
Nevermind, I see this is fixed by #193.

PR #199, opened 2016-07-22
Support inheriting a listening socket

djs55:
When starting a set of services together it is useful to be able to
pre-bind the listening sockets, to avoid races where a `connect` fails
because one of the components hasn't yet called `listen`.

This patch adds support for arguments of the form

  --arg fd:4

where '4' would be the listening socket file descriptor created by
the parent process.

### DIFF
--- a/src/conduit/datakit_conduit.ml
+++ b/src/conduit/datakit_conduit.ml
@@ -162,9 +162,10 @@ let accept_forever ~sandbox ~serviceid ~make_root url =
          Log.info (fun f -> f "Accepting connections on named pipe %s" url);
          Named_pipe.accept_forever url (Unix.handle ~make_root)
        end else if String.is_prefix ~affix:"fd:" url then begin
-         let i = String.with_range ~first:3 ~len:(String.length url - 3) url in
-         ( try Lwt.return (int_of_string i)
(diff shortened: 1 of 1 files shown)


src/conduit/datakit_conduit.ml
```diff
@@ -153,6 +161,14 @@ let accept_forever ~sandbox ~serviceid ~make_root url =
        if Astring.String.is_prefix ~affix:"\\\\" url then begin
          Log.info (fun f -> f "Accepting connections on named pipe %s" url);
          Named_pipe.accept_forever url (Unix.handle ~make_root)
+       end else if String.is_prefix ~affix:"fd:" url then begin
+         let i = String.with_range ~first:3 ~len:(String.length url - 3) url in
```
talex5: (note: you don't need `~len` here; taking the rest of the string is the default)

src/conduit/datakit_conduit.ml
```diff
@@ -153,6 +161,14 @@ let accept_forever ~sandbox ~serviceid ~make_root url =
...
+         ( try Lwt.return (int_of_string i)
```
talex5: (also: `String.to_int` has saner behaviour, returning an `option`)

talex5:
LGTM.

PR #200, opened 2016-07-24
More robustness fixes

samoht:


samoht:
Merging to ease the deployment on hub, but review still very much appreciated.

Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN opam depext lwt && opam install lwt alcotest conf-libev
 
 # cache opam install of dependencies
 COPY opam /home/opam/src/datakit/opam
-RUN opam pin add datakit.dev /home/opam/src/datakit -n
```
talex5: Not sure we want the `#HEAD` here. That means we only test committed changes (the `-k` makes sense though).

talex5:
Don't think I can review this - don't know enough about it at the moment. Maybe we should schedule a meeting for a code-walkthrough?

PR #205, opened 2016-07-26
Revert 90db49e27bc1a0a6adf021733c76e49877c0bec9

samoht:
Pinning in mixed mode make `datakit --version` report "%%VERSION%%"

/cc @dbuenzli and @altgr

I am not sure to understand why `-k git` does not copy the `.git` folder around, which then confuses the heuristic of `topkg` to detect if there is a `vcs` or not.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,8 +9,8 @@ RUN opam pin add datakit.dev /home/opam/src/datakit -n
 RUN opam depext datakit && opam install datakit --deps
 
 COPY . /home/opam/src/datakit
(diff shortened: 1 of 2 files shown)


dbuenzli:
> I am not sure to understand why -k git does not copy the .git folder around, which then confuses the heuristic of topkg to detect if there is a vcs or not.

It doesn't *confuses* topkg, topkg simply can't get the version number and leaves the version undefined. I reported this a long time ago in https://github.com/ocaml/opam/issues/2141. Just don't use `mixed-mode` it's a broken mode anways.

PR #207, opened 2016-07-26
On SIGTERM, on non-Win32, when we say we're going to exit, exit

djs55:
Previously we logged that we were going to exit but we forgot to
actually follow through.


talex5:
LGTM, once the test failure is fixed.

samoht:
I've restarted CircleCI with a fresh cache.

djs55:
I also rebased against latest master.

PR #208, opened 2016-07-26
Update the Golang bindings

dave-tucker:
This PR brings in changes from Docker for Mac and Windows to the Golang Datakit bindings.
This is necessary for us to switch to using the bindings upstream instead of our internal package

### DIFF
--- a/api/go-datakit/config.go
+++ b/api/go-datakit/config.go
@@ -114,6 +114,27 @@ func (r *Record) fillInDefault(path []string, value string) error {
 	return t.Commit(ctx, fmt.Sprintf("fill-in default for %s", path))
 }
 
+func (r *Record) SetMultiple(description string, fields []*StringField, values []string) error {
+	if len(fields) != len(values) {
+		return fmt.Errorf("Length of fields and values is not equal")
+	}
+	ctx := context.Background()
+	t, err := NewTransaction(ctx, r.client, "master", description)
+	if err != nil {
+		return err
+	}
+	for i, k := range fields {
+		p := append(r.path, k.path...)
+		v := values[i]
+		log.Printf("Setting value in store: %#v=%s\n", p, v)
+		err = t.Write(ctx, p, v)
+		if err != nil {
+			return err
+		}
+	}
+	return t.Commit(ctx, "Set multiple fields")
+}
+
 // StringField is a key which is associated with a string value
 type StringField struct {
 	path         []string
--- a/api/go-datakit/config.go
+++ b/api/go-datakit/config.go
@@ -19,7 +19,7 @@ type Record struct {
 	path     []string // directory inside the store
 	version  Version
 	schemaF  *IntField
-	fields   []*StringField // registered fields, for schema upgrades
+	fields   []*StringRefField // registered fields, for schema upgrades
 	branch   string
 	w        *Watch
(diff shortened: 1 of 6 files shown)


api/go-datakit/config.go
```diff
@@ -178,14 +226,48 @@ func (f *Record) StringField(key string, value string) *StringField {
...
-	raw        *StringField
+	raw        *StringRefField
 	defaultInt int
 }
 
 // Get retrieves the current value of the key
 func (f *IntField) Get() (int, Version) {
-	value64, err := strconv.ParseInt(strings.TrimSpace(f.raw.value), 10, 0)
+	if f.raw.value == nil {
+		log.Printf("Key %s missing in database, defaulting value to to %t", strings.Join(f.raw.path, "/"), f.defaultInt)
```
talex5: Typo: `to to`.

api/go-datakit/config.go
```diff
@@ -202,18 +284,22 @@ func (f *IntField) HasChanged(version Version) bool {
...
-	raw         *StringField
+	raw         *StringRefField
 	defaultBool bool
 }
 
 // Get retrieves the current value of the key
 func (f *BoolField) Get() (bool, Version) {
-	value, err := strconv.ParseBool(strings.TrimSpace(f.raw.value))
+	if f.raw.value == nil {
+		log.Printf("Key %s missing in database, defaulting value to to %t", strings.Join(f.raw.path, "/"), f.defaultBool)
```
talex5: `to to`

api/go-datakit/transaction_test.go
```diff
@@ -24,7 +24,7 @@ func TestTransaction(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Transaction.Write failed: %v", err)
 	}
-	err = trans.Commit(ctx)
```
talex5: How did this work before?

dave-tucker: Signature was changed in 50d61ee but the tests weren't updated

samoht: we should probably run these tests in the CI ...

api/go-datakit/transaction.go
```diff
@@ -96,12 +96,15 @@ func (t *transaction) Write(ctx context.Context, path []string, value string) er
 	return nil
 }
 
-// Remove a path (which can either be a key or a directory).
+// Remove deletes a key within the transaction.
 func (t *transaction) Remove(ctx context.Context, path []string) error {
 	p := []string{"branch", t.fromBranch, "transactions", t.newBranch, "rw"}
-	err := t.client.Remove(ctx, path...)
+	for _, dir := range path[0 : len(path)-1] {
```
samoht: this is indeed probably wrong, we want `len(path)` here. Or just append `path` to `p`.

talex5: Agree. Dropping the last element of `path` makes no sense.

talex5: I think you can just do `range path` on its own here to get all the elements.

talex5:
This also explains why the bridge fails to handle PRs being deleted (the Go hooks try to delete the wrong directory). Though I'm not sure why we don't see a `Failed to Remove` message in the logs when that happens.

PR #221, opened 2016-08-12
Implement transaction diffs

samoht:
With that commit, it is now possible to get the diff between an open transaction and any existing commits. To do that, simply `cat diff/<commit-id>` in the transaction directory.

### DIFF
--- a/src/client/datakit_client_9p.ml
+++ b/src/client/datakit_client_9p.ml
@@ -394,6 +394,8 @@ module Make(P9p : Protocol_9p_client.S) = struct
     let tree t = { Tree.fs = t.fs; path = path t / "ro" }
     let message t = FS.read_all t.fs (path t / "msg") >|*= Cstruct.to_string
     let id t = t.id
+    let pp ppf t = Fmt.string ppf t.id
+    let compare x y = String.compare x.id y.id
 
     let parents t =
       FS.read_all t.fs (path t / "parents") >|*= fun data ->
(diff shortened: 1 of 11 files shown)


src/client/datakit_client_9p.ml
```diff
@@ -60,6 +66,21 @@ let abort_if_off switch fn =
...
+      match String.cut ~sep:" " line with
+      | None            -> err "missing space"; acc
+      | Some (op, path) ->
+        match Datakit_path.of_string path with
+        | Error e -> err e; acc
+        | Ok path -> match op with
+          | "+" -> (`Added path  ) :: acc
+          | "-" -> (`Removed path) :: acc
+          | "*" -> (`Updated path) :: acc
+          | e   -> err e ; acc
```
talex5: Why do we continue and return successfully after an error?

src/ivfs/ivfs_tree.ml
```diff
@@ -228,6 +225,93 @@ module Make (Store : STORE) = struct
...
+    let lookup_file t n =
+      lookup t n >|= function
+      | `File f -> f
+      | _        -> failwith "lookup_file"
+
+    let iter t fn =
+      let rec aux = function
+        | []            -> Lwt.return_unit
+        | (t, path)::tl ->
+          ls t >>= fun childs ->
```
talex5: You can use `map` instead of `ls` here to avoid looking things up twice.

PR #222, opened 2016-08-12
Diff fixes and expose more functions in the client API

samoht:


src/client/datakit_S.mli
```diff
@@ -253,7 +257,13 @@ module type CLIENT = sig
...
         avoid the warning. *)
 
-      val diff: t -> Commit.t -> Datakit_path.t diff list or_error Lwt.t
+    val transaction: t -> Transaction.t or_error Lwt.t
+    (** [transaction t] creates a new transaction on top of the branch
+        [t]. Must be closed with {!Transaction.abort} or
+        {!Transaction.commit}. Use {!with_transaction} to not have to
+        worry about resource leaks.. *)
+
+    val diff: t -> Commit.t -> Datakit_path.t diff list or_error Lwt.t
```
talex5: Not keen on this one. Don't see any way to make it non-racey.

samoht: when the transaction is in control of one client only, this is pretty useful. How would you do it otherwise?

But yes, maybe that's fine if we just have a way to create a transaction from any commit and merge it back in a branch.

talex5: The one on Transaction is fine, but this one is on Branch. I've made a PR to move it here: https://github.com/docker/datakit/pull/224

Issue #237, opened 2016-08-22
GitHub bridge fails to remove old commit


talex5:
Here, commit 123 is no longer referenced, but the bridge does not remove it from DataKit:

```
GitHub:
{("a",
  {("a",
    prs=[n=0;head=456;title="PR#0";open];
    commits=[123->[[a/a:123:ci/datakit/test, description=Testing..., pending];
                   [a/a:123:ci/circleci, description=Testing..., failure]]])
   ("b", prs=[];commits=[])})
 ("b", {("a", prs=[];commits=[]) ("b", prs=[];commits=[])})}
DataKit:
{("a",
  {("a",
    prs=[n=0;head=456;title="PR#0";open];
    commits=[123->[[a/a:123:ci/circleci, description=Testing..., success];
                   [a/a:123:ci/datakit/test, description=Testing..., success]]])
   ("b", prs=[];commits=[])})
 ("b", {("a", prs=[];commits=[]) ("b", prs=[];commits=[])})}
```

It looks like we only remove commits that are pointed to by a closed PR. But no PR (open or closed) now has this head (because the head changed and the PR was closed before the bridge noticed).

(test case: https://github.com/talex5/datakit/commit/60918077809ff4c710844fa2e65aa09376c72fa6)

PR #238, opened 2016-08-23
Bridge tests (continued) + fixes

samoht:
Sorry @talex5 my rebases rewrote the author of the test commits. I can try to fix that on the next rebase.

The good new is that it seems that more tests now pass (e.g #237 is fixed). The bad news is that I haven't tried to uncomment the rest of the TODOs in the tests :-)


samoht:
All green, merging!

PR #241, opened 2016-08-25
github: allow the Github API calls to fail

samoht:


talex5:
I suspect the reason we didn't spot this earlier is that we use `Lwt.join` instead of `Lwt.choose` to monitor all the threads. That means we'll only see an error if *all* the threads fail.

PR #242, opened 2016-08-25
Better references synchronisation for the Github bridge

samoht:


tests/test_github.ml
```diff
@@ -818,37 +838,99 @@ let test_contexts = [|
...
+let random_description ~random = random_choice ~random test_descriptions
+
+let random_state ~random = random_choice ~random test_state
+
+let random_refs ~random ~repo ~old_refs =
+  test_names
+  |> Array.to_list
+  |> List.map (fun name ->
+      if Random.State.bool random then (
+        match List.find (fun r -> r.Ref.name = name) old_refs with
```
talex5: I don't think you need `old_refs` here. GitHub is fine with deleting existing refs.

PR #256, opened 2016-09-09
Record local timestamps, not microseconds since start

talex5:
This makes it easier to correlate with other logs.

Before:
```
+5817us    conduit [INFO] Waiting for connections on Unix socket tcp://127.0.0.1:5640
```

After:
```
2016-09-09 15:32    conduit [INFO] Waiting for connections on Unix socket tcp://127.0.0.1:5640
```

Does anyone need the monotonic time for something?

Might be nice to include ms and tz offset from GMT, but Unix API doesn't make this very easy.

### DIFF
--- a/_tags
+++ b/_tags
@@ -19,7 +19,7 @@ true: package(bytes lwt astring logs result cstruct fmt rresult)
 <src/conduit/*>: package(protocol-9p.unix)
 
 <src/log>: include
-<src/log/*>: package(asl win-eventlog cmdliner mtime.os logs.cli)
+<src/log/*>: package(asl win-eventlog cmdliner logs.cli)
 
 <src/irmin-io> : include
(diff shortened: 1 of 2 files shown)


dsheets:
dbuenzli/mtime#14 gives you a way to access the system-relative mtime for log correlation purposes. If you want precise event ordering across different processes on the same machine in the same boot session, it contains the functionality you want. I don't think it is particularly confusing or needs further work but I also haven't put in the effort since the original patch to get it into upstream. Docker for Mac uses the patch for file system event logging.

talex5:
That won't work across machines, though, right? e.g. if GitHub says it sent me a webhook message at `2016-09-09 15:01:16` then I can't find the corresponding place in the DataKit logs.

dsheets:
No, it won't work across machines. The best you can do is record a wall-clock and monotonic clock pair to provide known moments for cross-machine correlation (but still have nice monotonicity properties for local events).

PR #260, opened 2016-09-16
Fix bridge event loop when not listening for 9p connections

### LINKED ISSUE #259: GitHub bridge doesn't work without listeners
I broke this in #253 :-(

If there are listeners, then `accept_connections` never returns (and the error reporting behaviour is unclear).
If there aren't, then it returns immediately.

So, returning is sometimes an error and sometimes not. Needs a better fix.

talex5:
Was broken by https://github.com/docker/datakit/pull/253.
Fixes #259.


talex5:
Not sure what's up with AppVeyor, but considering that another PR that just added a README also broke it, ignoring it.

PR #262, opened 2016-09-18
Fix Appveyor

dave-tucker:
This should work based on https://github.com/docker/vpnkit/commit/1669eb1925ee43afd5a46188a43bf030bf34e01d and https://github.com/mirage/ocaml-github/pull/147 ... and full credit goes to @MagnusS for finding the root cause!

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,7 +8,7 @@ environment:
 
 install:
   - appveyor DownloadFile https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/appveyor-opam.sh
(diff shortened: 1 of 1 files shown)


talex5:
Yay!! Does this fix need to be upstreamed too?

https://github.com/ocaml/ocaml-ci-scripts

Looks like it has the old version in its example: https://github.com/ocaml/ocaml-ci-scripts/blob/master/README-appveyor.md

dave-tucker:
I wasn't aware there was an upstream! Had a chat with @MagnusS this morning and he'll take care of the upstream fix

MagnusS:
PR for upstream: https://github.com/ocaml/ocaml-ci-scripts/pull/128

PR #270, opened 2016-09-23
github: trim the build status description before projecting it in the FS

samoht:


samoht:
CircleCI is still broken (the OSX VM does not have the right version of XCode), so merging.

PR #272, opened 2016-09-23
Speed-up start-up time

samoht:
- Fix pruning (so there is less files to scan on startup)
- Do less 9p operations when walking a tree

### DIFF
--- a/bridge/github/src/datakit_github.ml
+++ b/bridge/github/src/datakit_github.ml
@@ -668,37 +668,36 @@ module Conv (DK: Datakit_S.CLIENT) = struct
     | Error _ -> false
     | Ok b    -> b
 
+  let safe_exists_file (E ((module Tree), tree)) file =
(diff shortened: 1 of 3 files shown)


bridge/github/src/datakit_github.ml
```diff
@@ -668,37 +668,36 @@ module Conv (DK: Datakit_S.CLIENT) = struct
...
-        else
-          Lwt.return child
+    let rec aux acc = function
+      | [] -> Lwt.return acc
+      | context :: todo ->
+        match Datakit_path.of_steps context with
+        | Error e -> Log.err (fun l -> l "%s" e); aux acc todo
+        | Ok ctx  ->
+          let dir = root /@ ctx in
+          safe_read_dir tree dir >>= fun childs ->
```
talex5: Maybe we should change `read_dir` to say whether each item is a file or a directory. The server provides this information in the 9p protocol.

PR #278, opened 2016-09-28
Update/fix META file requires

djs55:
This fixes `#require` in the top-level.

### DIFF
--- a/pkg/META
+++ b/pkg/META
@@ -6,7 +6,7 @@ package "vfs" (
  directory = "vfs"
  description = "Datakit's VFS description"
  version = "%%VERSION%%"
- requires = ""
+ requires = "fmt astring lwt cstruct logs result"
  archive(byte) = "vfs.cma"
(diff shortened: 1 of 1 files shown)


pkg/META
```diff
@@ -6,7 +6,7 @@ package "vfs" (
  directory = "vfs"
  description = "Datakit's VFS description"
  version = "%%VERSION%%"
- requires = ""
+ requires = "fmt astring lwt cstruct logs unix sexplib result"
```
talex5: Why does vfs depend on Unix?

djs55: Not sure, I got this from

djs55: Maybe it's a false positive of some kind? When I comment code out, it disappears after I remove all references to `Cstruct`.

djs55: I've removed `unix` and `sexplib` and it seems to still work in `utop`. I clearly don't understand what `ocamlobjinfo` does.

PR #281, opened 2016-09-30
Expose the datakit API over a very basic REST API.

samoht:
/cc @dgageot 

I testing a few more things and I'll merge, so new binary available shortly.

### DIFF
--- a/src/bin/main.ml
+++ b/src/bin/main.ml
@@ -57,14 +57,16 @@ module Git_fs_store = struct
     Irmin.Private.Watch.set_listen_dir_hook (Irmin_watcher_polling.hook 1.)
   )
 
-  let repo path =
+  let repo ~sandbox path =
+    let path = (if sandbox then "." else "") ^ path in
     let config = Irmin_git.config ~root:path ~bare:true () in
     Store.Repo.create config
 
-  let connect path =
+  let connect ~sandbox path =
     Lazy.force listener;
-    Log.debug (fun l -> l "Using Git-format store %S" path);
-    repo path >|= fun repo ->
+    let s = if sandbox then " (sandboxed)" else "" in
+    Log.debug (fun l -> l "Using Git-format store %S%s" path s);
+    repo ~sandbox path >|= fun repo ->
     fun () -> Filesystem.create make_task repo
 end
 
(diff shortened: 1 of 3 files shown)


src/bin/main.ml
```diff
@@ -110,15 +146,22 @@ let start urls sandbox git =
...
+      | None      -> In_memory_store.connect ()
+      | Some path ->
+        let prefix = if sandbox then "." else "" in
+        Git_fs_store.connect (prefix ^ path)
+    end >>= fun make_root ->
+    Lwt_list.iter_p
+      (Datakit_conduit.accept_forever ~make_root ~sandbox ~serviceid)
+      listen_9p
+  in
+  Lwt.join [ serve_http; serve_9p ]
```
talex5: Using `Lwt.join` hides errors, as we discovered before. We should just make a list of the threads we actually want to run and use `Lwt.choose`.

src/bin/main.ml
```diff
@@ -92,7 +92,43 @@ let set_signal_if_supported signal handler =
...
+    | Some _ -> uri in
+  let port, uri = match Uri.port uri with
+    | None   -> 8080, Uri.with_port uri (Some 8080)
+    | Some p -> p, uri in
+  let mode = `TCP (`Port port) in
+  Logs.info (fun f -> f "daemon: %s" (Uri.to_string uri));
+  Printf.printf "Server starting on port %d.\n%!" port;
+  begin match git with
+    | None -> (* in-memory store *)
+      let module HTTP = HTTP(In_memory_store.Store) in
```
talex5: Why are we creating the store inside the HTTP code? Shouldn't the store be shared?

samoht: The store handle is shared by all the HTTP requests to read and write stuff. Clearly not ideal. Also, we pick a default branch (in this case `master`) to display when accessing the top-level  paths (such as`/read/<path>` which is a short-cut to `/tree/master/read/<path>`.

Obviously that API has a lot of [issues](https://github.com/mirage/irmin/issues?q=is%3Aopen+is%3Aissue+label%3A%22REST+API%22) but @dgageot wanted to experiment a bit with it so I resurrected it.

src/bin/main.ml
samoht: ok, I'll amend the PR.

samoht: done.

PR #282, opened 2016-10-02
Add more to datakit-client

djs55:
- Add `datakit-client.fs9p` for applications which want to expose a filesystem created with `datakit-client.vfs` over 9P (e.g. [docker/vpnkit#123])
- Fix the `requires` line of `datakit-client.vfs` (changes were not copied from the other META file)
- Both `.fs9p` sub-packages should depend on `protocol-9p` rather than `protocol-9p.unix` since `Fs9p` is functorised over `V1_LWT.FLOW` which makes it Mirage Unikernel-friendly

### DIFF
--- a/pkg/META
+++ b/pkg/META
@@ -30,7 +30,7 @@ package "fs9p" (
  directory = "fs9p"
  description = "Datakit's VFS to 9p layer"
  version = "%%VERSION%%"
- requires = "protocol-9p.unix datakit.vfs"
+ requires = "protocol-9p datakit.vfs"
  archive(byte) = "fs9p.cma"
  archive(native) = "fs9p.cmxa"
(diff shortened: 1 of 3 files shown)


talex5:
`datakit-client.vfs`? How did all this server stuff get into the client package?

djs55:
It wasn't me! :) I agree it's odd ... perhaps the word "client" should be removed from the package name

I don't particularly mind where the code lives (datakit-foo.bar, protocol-9p.foo, mirage-vfs) ... anywhere as long as I can link to it rather than cut 'n paste it :)

djs55:
Hm, perhaps this overlaps with the unmerged https://github.com/docker/datakit/pull/280

samoht:
Yes, this should be fixed by #280, so closing.

Issue #298, opened 2016-10-10
Datakit-client requires datakit-client to be installed before compiling?


avsm:
```
#=== ERROR while installing datakit-client.dev~mirage =========================#
# opam-version 1.2.2
# os           darwin
# command      ocaml pkg/pkg.ml build --pinned false -n datakit-client
# path         /Users/avsm/.opam/system/build/datakit-client.dev~mirage
# compiler     system (4.03.0)
# exit-code    1
# env-file     /Users/avsm/.opam/system/build/datakit-client.dev~mirage/datakit-client-29188-4f1270.env
# stdout-file  /Users/avsm/.opam/system/build/datakit-client.dev~mirage/datakit-client-29188-4f1270.out
# stderr-file  /Users/avsm/.opam/system/build/datakit-client.dev~mirage/datakit-client-29188-4f1270.err
### stdout ###
# [...]
# ocamlfind ocamldep -package 'bytes lwt astring logs result cstruct fmt rresult' -modules src/version.ml > src/version.ml.depends
# ocamlfind ocamlc -c -g -bin-annot -safe-string -package 'bytes lwt astring logs result cstruct fmt rresult' -warn-error +1..49 -w A-3-4-41-44 -I src -I src/datakit -o src/version.cmo src/version.ml
# ocamlfind ocamlc -c -g -bin-annot -safe-string -package 'cmdliner fmt.cli fmt.tty logs.fmt' -package 'protocol-9p uri' -package 'bytes lwt astring logs result cstruct fmt rresult' -warn-error +1..49 -w A-3-4-41-44 -I src/datakit-client -I src -I src/datakit -o src/datakit-client/mount.cmo src/datakit-client/mount.ml
# ocamlfind ocamlopt -c -g -bin-annot -safe-string -package 'bytes lwt astring logs result cstruct fmt rresult' -warn-error +1..49 -w A-3-4-41-44 -I src -I src/datakit -o src/version.cmx src/version.ml
# ocamlfind ocamlopt -c -g -bin-annot -safe-string -package 'cmdliner fmt.cli fmt.tty logs.fmt' -package 'protocol-9p uri' -package 'bytes lwt astring logs result cstruct fmt rresult' -warn-error +1..49 -w A-3-4-41-44 -I src/datakit-client -I src -I src/datakit -o src/datakit-client/mount.cmx src/datakit-client/mount.ml
[...]

PR #299, opened 2016-10-10
do not compile example program as part of tests

avsm:
This requires datakit-client to be installed but is currently part
of the datakit-client build. See #298


pkg/pkg.ml
```diff
@@ -38,7 +38,6 @@ let () =
       Pkg.lib   "src/datakit-client/datakit_S.mli";
       Pkg.lib   "src/datakit-client/datakit_S.cmi";
       Pkg.bin   "src/datakit-client/mount" ~dst:"datakit-mount" ;
-      Pkg.test  "examples/ocaml-client/example" ~run:false ;
```
samoht: Can you move this to `"opam"` switch instead of removing it completely?

PR #304, opened 2016-10-28
ci: allow configuring CI name

talex5:
This is displayed in the web page header, and is also used to make the
cookie name unique in case you're running multiple CIs on one host.

Also, configure the web UI with code, not command-line options. Allows
type-checking configuration before deployment and makes container image
more self-contained.

Signed-off-by: Thomas Leonard thomas.leonard@docker.com


talex5:
Superceeded by #306.

PR #306, opened 2016-10-31
ci: API cleanups

talex5:
* `--state-repo` and `--dashboard` options have been replaced with settings in the config script. This also avoids needing to check that the dashboard configuration matches the set of projects, since they're now specified in the same place.
* If there is no dashboard configuration, we now default to showing `master`.
* The CI is configured with a name (e.g. `example-ci`). This makes it easy to see from the web page header which CI you're looking at, and also avoids a problem with clashing session cookies if you deploy multiple CIs on different ports.


talex5:
(note: Travis failure will be fixed by https://github.com/ocaml/opam-repository/pull/7718)

talex5:
Merging despite CI failure, as the same commit passed the same tests in my fork (the error is a race installing Irmin).

PR #310, opened 2016-11-01
Remove ncurses-dev from Dockerfiles

talex5:
This isn't needed any longer. Can just install `conf-libev` now.

See: https://github.com/docker/datakit/pull/302#discussion_r85918731


talex5:
Docker builds failed with
```
# The following recquired C libraries are missing: libev.
```

Need to set up a proper CI for DataKit, that tests the builds *before* they get to master...

avsm:
That should be brought in by `conf-libev`'s depext.

avsm:
Which is already present...hmm

PR #311, opened 2016-11-02
3rd iteration of the GitHub bridge

samoht:
Previously, the bridge was using two DataKit branches:

- a "private" branch which will mirror the GitHub state, populated on startup by GitHub API calls, and updated regularly by a separate webhook listener directly connected to DataKit.

- a "public" branch, regularly merged with the private one, where users could see the current state of GitHub and also change some parts of the state (for instance, update the build status of a given commit) to trigger appropriate GitHub API calls.

Using that scheme: conflicts could happen, (as webhooks could arrive at the same time as a user request to change some state), the merge strategy is handled by the DataKit server (and hence, it doesn't understand the invariants that should be kept in the GH state). Moreover, as the bridge doesn't have any handle on both the incoming webhooks and the user requests, it is pretty hard for it to keep a consistent view of the world without loosing events (either from GH or from the users).  Finally, any conflict/error could lead to surprising results ("oh, the GH bridge reverted the title of that PR to something old!")  and was a sometimes unreliable.

This (rather large) patches refactor the existing code to simplify
that process.

- There is now only one branch in DataKit, where users can see the state of GitHub and request changes (the equivalent of the previous "public" branch).

- The "private" branch now only exists in the bridge memory, and is rebuilt from scratch on every restart. Webhooks events are kept in a queue and read/applied when needed (instead of arriving in
  asynchronous order at any point in time). Conflict are handled by letting the user specifying extended capabilities/ownership on every GH resources. This let for instance users the ability to only modify the build status starting by a given prefix, and nothing else.

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -3,8 +3,9 @@ PKG alcotest rresult github astring fmt irmin logs mtime.os
 PKG camlzip irmin.mem irmin-watcher
 PKG conduit.lwt-unix hvsock named-pipe
 PKG asl win-eventlog github-hooks
+PKG datakit-client datakit-server.vfs datakit-github
 
 S src/**
 S tests/
-S bridge/github/**
+S bridge/github/*
 B _build/**
--- a/Dockerfile.github
+++ b/Dockerfile.github
@@ -1,6 +1,7 @@
 FROM docker/datakit:server
 
 COPY datakit-github.opam /home/opam/src/datakit/datakit-github.opam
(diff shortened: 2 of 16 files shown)


talex5:
Note: CI failures here seem to be genuine:

```
- File "bridge/github/datakit_github.ml", line 1071, characters 25-27:
- Error: Unbound constructor Ok
```

talex5:
Rebasing on master would fix a couple more of the CI failures.

PR #313, opened 2016-11-02
Fix building with libev

talex5:
I broke this in #310.

### DIFF
--- a/Dockerfile.client
+++ b/Dockerfile.client
@@ -1,6 +1,6 @@
 FROM ocaml/opam:alpine
 
-RUN opam depext lwt && opam install lwt alcotest conf-libev
+RUN opam depext -ui lwt alcotest conf-libev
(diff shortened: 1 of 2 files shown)


talex5:
@samoht: bridge closed my PR :-(

Dockerfile.client
```diff
@@ -1,6 +1,6 @@
 FROM ocaml/opam:alpine
 
-RUN opam depext lwt && opam install lwt alcotest conf-libev
+RUN opam depext -i lwt alcotest conf-libev
```
avsm: This should be `-ui` to be reliable in the face of upstream revision updates (it triggers an `apk update` before attempting to install)

talex5:
Only build failure is the bridge, due to an unrelated problem with build ordering, so merging.

PR #319, opened 2016-11-03
update opam metadata

avsm:
- enforce ocaml version 4.03 (datakit-ci uses `result` unconditionally)
- datakit-github depends on github>2.1.0 or higher


talex5:
Note: Travis is still testing the client with 4.02. We should either remove that test, or not restrict the client so much.

avsm:
Supporting the client for 4.02 is probably quite sensible; I'll amend the PR.

talex5:
Hmm, looks like most of the Dockerfiles assumed 4.02. We should probably update them. They're using the default OCaml version from the `ocaml/opam:alpine` base image.

Also, looks like some Travis pins need removing.

avsm:
1f39deef0c6b123df937da967ac46b03e9766e8e remove the travis pins
5910038ee57e830dfb457feb909d530792d10f99 specifies the ocaml versions as 4.03

talex5:
(failure of `datakit` and `github` is expected because they build against the base images on hub - hub will fail for the same reason, and @samoht will trigger a rebuild manually...)

talex5:
Update needed for AppVeyor too?

samoht:
I am not sure why this PR is needed. I seem to be able to compile datakit on 4.02 without too much trouble.

samoht:
Actually it seems that only datakit-ci needs 4.03. @avsm could you update your PR to only add the constraint for that package? Thanks!

PR #327, opened 2016-11-07
Catch top level Lwt.async exceptions in datakit and datakit-github

samoht:
When such exception is caught, just log it and continue instead of failing.

### DIFF
--- a/bridge/github/main.ml
+++ b/bridge/github/main.ml
@@ -79,6 +79,11 @@ let ( >>~ ) x f =
   | Error e -> Lwt.fail_with @@ Fmt.strf "%a" DK.pp_error e
   | Ok t    -> f t
 
+let () =
+  Lwt.async_exception_hook := (fun exn ->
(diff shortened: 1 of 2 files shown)


bridge/github/main.ml
```diff
@@ -79,6 +79,16 @@ let ( >>~ ) x f =
   | Error e -> Lwt.fail_with @@ Fmt.strf "%a" DK.pp_error e
   | Ok t    -> f t
 
+let () =
+  Lwt.async_exception_hook := (function
+      | Unix.Unix_error (error, func, arg) ->
+        Logs.warn (fun m ->
+            m  "Client connection error %s: %s(%S)"
```
talex5: This seems like a potentially misleading error. There's no particular reason to think this is related to client connections. Might make more sense to define a wrapper around `Lwt.async` that takes a context message to print.

samoht:
Updated.

PR #328, opened 2016-11-07
ci: add GitHub-based login to web UI

talex5:
TODO:

- [x] Add some kind of policy to say which GitHub users should be allowed to use the CI


talex5:
Not sure if access policy works with private repositories. Will check once I get approval from test org.

PR #329, opened 2016-11-07
ci: let CI binaries specify custom CLI term info

avsm:
This avoids them all being named "DataKitCI" and also allows
individual CI binaries to specify version and description information
for their CLIs.

Signed-off-by: Anil Madhavapeddy <anil@recoil.org>

### DIFF
--- a/ci/src/cI_main.ml
+++ b/ci/src/cI_main.ml
@@ -87,7 +87,9 @@ let canaries =
   in
   Arg.(value (opt_all CI_target.Full.arg [] doc))
 
-let run config =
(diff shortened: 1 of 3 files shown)


talex5:
Hmm. There's already a `name` in `Web.config`. Would it be better to reuse that?

avsm:
We could reuse `Web.config` for the default term, but for the but I'd like to match the name of the binary in my usecase, and only the build system knows what it'll eventually be called (and the binary name is probably less user friendly than the Web.config version)

talex5:
OK. I need to clean up the whole `Main` thing at some point...

PR #331, opened 2016-11-07
github: normalise the Commit and Ref modules

avsm:
Make the types private to ease the process of finding places
where they are allocated, and ensure that they are normalised
using `Astring.String.trim` to get rid of whitespace or errant
newlines.

### DIFF
--- a/bridge/github/datakit_github.ml
+++ b/bridge/github/datakit_github.ml
@@ -47,6 +47,11 @@ let pp_path = Fmt.(list ~sep:(unit "/") string)
 module Repo = struct
 
   type t = { user: string; repo: string }
+  let create ~user ~repo =
+   let user = String.trim user in
+   let repo = String.trim repo in
+   { user; repo }
+
   let pp ppf t = Fmt.pf ppf "%s/%s" t.user t.repo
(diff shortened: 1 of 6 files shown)


avsm:
I'm still testing this one locally, here for review

bridge/github/datakit_github.ml
```diff
@@ -95,6 +95,7 @@ module Commit = struct
 
   type t = { repo: Repo.t; id : string }
 
+  let create repo id = {repo; id = String.trim id }
```
talex5: I think it would make more sense to reject invalid arguments rather than silently trying to fix them. It would be very confusing for a name to work in some places and not others.

samoht: Adding or not a newline at the end of a file should not break the bridge. I think `String.trim` is sensible here, as it is already done for `PR.create` where the description is truncated to fit GH length constraints.

avsm: We could also run this in a future patch in an "invariants" mode where it issues a warning with the original callsite if `id <> String.trim id`. For now, this makes the interface more defensive, as @samoth notes

samoht:
I've pushed some changes to your branch, now all green so merging.

PR #332, opened 2016-11-08
retry named pipe connection on failure

simonferquel:
Tentative to fix Named pipe accept problems with Pinata

### DIFF
--- a/src/datakit_conduit.ml
+++ b/src/datakit_conduit.ml
@@ -140,7 +140,7 @@ module Named_pipe = struct
     Named_pipe_lwt.Server.connect p >>= function
     | false ->
       Log.err (fun f -> f "Named-pipe connection failed on %s" path);
-      Lwt.return ()
+      Lwt_unix.sleep 0.1 >>= fun () -> accept_forever ?backlog path callback;
     | true ->
       let _ = (* background thread *)
         let fd = Named_pipe_lwt.Server.to_fd p in


talex5:
@djs55 I don't understand how this is supposed to work. What does the bool returned by `connect` mean, and how should it be handled? The docs just say:

```
val connect: t -> bool Lwt.t
(** Connect blocks until a client connects to this named pipe *)
```

(also, there's a `let _ = ` a few lines below, which will probably make debugging problems harder than it needs to be by suppressing any exceptions that get thrown...)

simonferquel:
From the existing code, I would say that the Boolean indicates success or failure of connect, but I might be wrong.
My PR just reattempt the connection (in case the named pipe has not yet been freed by a previous instance). (for now I just made the PR, so I can get a build - for whatever reason I can't build datakit locally)

djs55:
I believed `true` means connected and `false` means something went wrong (I'm not sure why I didn't throw `Unix_error` in that case). I probably copied the style of the C function:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365146(v=vs.85).aspx

samoht:
I am working on an improved version of the named-pipe binding: https://github.com/mirage/ocaml-named-pipe/pull/14 which fixes the API too (and add some debugging statements).

I'll update the code in that PR to use this when this is ready (should be soon).

simonferquel:
This project fixes the issue on pinata. However something since last release seem to make it much more verbose in term of logging (every read is logged)

samoht:
Closing as #333 is not in.

PR #333, opened 2016-11-09
Improve named pipes

samoht:
Use my fork of ocaml-named-pipe for testing. Error handling has been cleaned up (so we should know why it is failing when it fails) and continue to accept when getting client errors. Subsume #332


talex5:
Genuine-looking CI failure:

```
 File "src/datakit_conduit.ml", line 149, characters 30-32:
 Error: This variant pattern is expected to have type bool
        The constructor () does not belong to type bool
```

PR #350, opened 2016-11-18
ci: show status after list of logs

talex5:
Having it at the top was confusing people, because often the last log
succeeded, but then something without a log failed.

Don't show the status at all if it's the same as the last step's title
(common for pending tasks).

Indicate the state of each step with an icon next to it, rather than
by subtle changes to the colour and shading of the rebuild button.


talex5:
CircleCI failed with:
```
- File "src/datakit/ivfs.ml", line 859, characters 35-47:
- Error: Unbound module Vfs.Logs
```
But that has nothing to do with this patch.

talex5:
`master` also failed a rebuild in the same way, so merging.

Issue #337, opened 2016-11-11
Extracting the details of the queue length from DatakitCI would be helpful for capacity planning


samoht:

talex5:
You can now use the Prometheus query `DataKitCI_pool_qlen` to see all the queue lengths.

PR #352, opened 2016-11-22
ci: add Prometheus metric reporting

talex5:
Initially, the metrics reported are:

- Total number of builds started
- Total number of builds that succeeded
- Total number of builds that failed
- Total number of builds that raised an exception
- Number of attempted connections to DataKit
- Number of failed attempts to connect to DataKit
- Number of status updates
- Number of notifications from DataKit
- Number of users waiting for a resource
- Time spent waiting for a resource
- Time spent using a resource
- Number of term evaluations
- HTTP requests to web UI
- HTTP responses from web UI
- HTTP requests currently being handled by the web UI
- Number of successful local login attempts
- Number of unsuccessful local login attempts
- Number of successful GitHub login attempts
- Number of unsuccessful GitHub login attempts
- Time to handle one web request

### DIFF
--- a/ci/skeleton/exampleCI.ml
+++ b/ci/skeleton/exampleCI.ml
@@ -19,12 +19,22 @@ let state_repo =
   None
   (* Some (Uri.of_string "https://github.com/my-org/my-project.logs") *)
 
+let metrics_token =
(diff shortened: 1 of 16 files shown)


talex5:
Closes #337

PR #357, opened 2016-11-24
Remove the --sandbox mode

samoht:
This was 1/ not well designed 2/ not secure and 3/ not used by anyone.


samoht:
I have checked with @djs55 and normally this should not have any impact on Docker for Mac and Windows.

PR #368, opened 2016-11-28
Add regular resync events and .dirty files

samoht:
This is configurable via `--resync-interval` one the command-liine (default is to do it every 10 minuutes, but that's probably too often).

I have also added support for `.dirty` files to allow the user to request a resync: just drop a `.dirty` in the sub-dir that you want to refresh (note: we only support prs, refs and repos at the moment)


samoht:
The `.dirty` files don't seem to work yet, so please don't merge.

PR #369, opened 2016-11-29
ci: allow hiding some arguments when logging commands

talex5:
This is useful to avoid logging secret tokens.


talex5:
Appveyor has been building for nearly an hour with no obvious sign of progress, so merging anyway.

PR #372, opened 2016-11-29
ci: do not percent-encode the ref names in the URL

samoht:


talex5:
Why are you changing the types? pct-encoding is just a web-UI issue.

samoht:
Why do we need to pass a `Datakit_path.t` around. A reference name is just a string.

ci/src/cI_web_templates.ml
```diff
@@ -51,9 +51,9 @@ let gh_target_url = function
   | `Ref r ->
     let { CI_projectID.user; project } = CI_github_hooks.Ref.project r in
     let id = CI_github_hooks.Ref.name r in
-    match Datakit_path.unwrap id with
-    | "tags" :: id -> Fmt.strf "https://github.com/%s/%s/releases/tag/%s" user project (String.concat ~sep:"/" id)
-    | _            -> Fmt.strf "https://github.com/%s/%s/tree/%a" user project Datakit_path.pp id
+    match String.cut ~sep:"tags/" id with
```
talex5: This is wrong. `tags` has to be the *first* component. Using strings is inviting these kinds of mistakes.

samoht: Would it be better if I use string list as in https://github.com/docker/datakit/blob/master/bridge/github/datakit_github.mli#L224 ?

samoht: Ideally we should converge all these types

talex5: (after some discussion)

- Using strings isn't great, because ref names do have structure and it's useful to match on it.

- `Datakit_path` could expose the type (`type t = private string list`). I didn't do this because I was considering making a more optimal representation (e.g. that makes appending fast), but maybe having to call `unwrap` is too annoying.

- We could abstract this with another type "git ref name" and let some library convert between this type and DataKit paths.

talex5:
Also, note that you still need to % encode the other characters in the URL, even if you exclude '/'.

samoht:
Closing, as this needs a larger refactoring to merge `CI_github_hook` and `Datakit_github`. Patches incoming.

PR #384, opened 2016-12-01
Refactor datakit-ci

samoht:
Apologies for the size of the patch, I tried to do the refactoring by small steps but the result is bigger that I was hoping for. Note: this is just the first step, I plan to continue the merge further, but having a review at this stage would already be useful.

The rational of the change is to make `datakit-ci` uses the data-structures exposed by `datakit-github` instead of redefining them. This has multiple advantages, the main ones being:

- More data are now available to the CI: the base branch for PRs is one crappy example, but anything that is added to the base type is now automatically exposed to the CI. This will be useful to expose creation date and other metadata.
- The 9p conversion is handled by `datakit-github` already so no need to re-encode everything again. This will make changing the directory hierarchy seemless (e.g. the CI won't care).
- `datakit-github` should expose more resources soon (e.g. issues and comments). This will make them available to the CI without big effort.


ci/src/cI_config.ml
```diff
@@ -1,26 +1,31 @@
...
 }
 
-let id_of_branch name =
-  `Ref (Datakit_path.of_string_exn ("heads/" ^ name))
+let id_of_branch repo name = `Ref (repo, "heads" :: String.cuts ~sep:"/" name)
 
 let project ~id ?(dashboards=["master"]) tests =
-  let id = CI_projectID.of_string_exn id in
+  let id = match Repo.of_string id with
+    | None   -> failwith ("project: " ^ id)
```
talex5: I don't think these errors are very clear. Might be worth giving `Repo.of_string` a `result` type. I can't think of any reason why you'd care that a repo name was invalid and not want to log the reason somewhere.

Failing that, `Utils.failf "Invalid repo ID %S" id` would be clearer.

ci/src/cI_config.ml
```diff
@@ -1,26 +1,31 @@
...
+    | Some r -> r
+  in
   let tests x = String.Map.of_list (tests x) in
-  let dashboards = CI_target.ID_Set.of_list (List.map id_of_branch dashboards) in
+  let dashboards =
+    CI_target.Set.of_list (List.map (id_of_branch id) dashboards)
+  in
```
talex5: Can we upgrade to this: https://en.wikipedia.org/wiki/Punched_card#IBM_96-column_punched_card_format ?

ci/src/cI_github_hooks.mli
```diff
@@ -1,80 +1,17 @@
...
-  val status : t -> Datakit_S.status_state option Lwt.t
-  val descr : t -> string option Lwt.t
-  val target_url : t -> Uri.t option Lwt.t
-end
+open Datakit_github
 
 module CI : sig
-  type t
-  val of_string : string -> t
+  type t = string list
```
talex5: Why is this no longer abstract? (doesn't matter much, but seems strange that we need to expose it now if we didn't before)

avsm:
would it be possible to release a version of datakit-ci before merging this patch? That would let the existing production CIs migration more easily at their own pace

ci/src/cI_engine.ml
```diff
@@ -106,67 +107,131 @@ let rec connect connect_dk =
...
+    Repo.Map.mapi (fun id make_terms ->
         let canaries =
           match canaries with
-          | None -> None
-          | Some canaries -> Some (CI_projectID.Map.find id canaries |> default CI_target.ID_Set.empty)
+          | None          -> None
+          | Some canaries ->
+            match Repo.Map.find id canaries with
+            | Some s -> Some s
+            | None   -> Some CI_target.Set.empty
```
talex5: You don't like my `default` function?

ci/src/cI_engine.ml
```diff
@@ -106,67 +107,131 @@ let rec connect connect_dk =
...
-  t.projects |> CI_projectID.Map.map (fun project -> project.open_prs, project.refs)
 
-let escape_ref path =
-  Uri.pct_encode ~scheme:"http" (String.concat ~sep:"/" (Datakit_path.unwrap path))
+let update_status t s =
+  t.dk >>= fun t ->
+  DK.branch t metadata_branch >>*= fun metadata ->
+  DK.Branch.with_transaction metadata (fun t ->
+      Conv.update_elt t (`Status s) >>= fun () ->
+      let message =
```
talex5: Note: I updated this message in #385

ci/src/cI_config.ml
samoht: Thanks, changed.

ci/src/cI_config.ml
samoht: Sorry I am using an editor with [hard-coded 80-column rule](https://www.emacswiki.org/emacs/EightyColumnRule) and it changes the color of too long lines :p 

I am trying to resist the urge to reformat the whole source-code, it's pretty hard :-) Anyway, I've reverted that change.

ci/src/cI_github_hooks.ml
```diff
@@ -1,293 +1,56 @@
...
-  type t = [ `PR of PR.t | `Ref of Ref.t ]
-
-  let dispatch p r = function
-    | `PR x -> p x
-    | `Ref x -> r x
-
-  let head = dispatch PR.head Ref.head
+  type t = string list
+  let circle_ci = ["ci"; "circleci"]
+  let datakit_ci x = "ci" :: "datakit" :: String.cuts ~sep:"/" x
```
samoht: I will revert that, that change got intertwined with the big rewrite, which is bad.

ci/src/cI_github_hooks.mli
```diff
@@ -1,80 +1,17 @@
...
-val set_state : t -> CI.t -> status:Datakit_S.status_state -> descr:string -> ?target_url:Uri.t -> Commit.t -> unit Lwt.t
-
-module Target : sig
-  type t = [ `PR of PR.t | `Ref of Ref.t ]
-
-  val head : t -> Commit.t
-end
-
 module Snapshot : sig
-  type t
```
samoht: Yes, but my plan was to add it back in the client library if really needed. Do you think that performance will become very bad without the cache? Did you add it because you measured bottleneck in the 9p layer?

ci/src/cI_engine.ml
samoht: changed back to use `default`

samoht:
Made the changes requested by @talex5 and I've continued the simplification of `CI_github_hooks`. It's now an empty shell, I'll see how I can remove the last 2 functions.

What is missing before merging:
- Need to tag a release to allow the current user to update
- Need to figure out if the 9p caching was needed and if yes, need to re-add it but maybe in client library.

ci/src/cI_github_hooks.mli
talex5: Well, we know that scanning a project is slow (I think you've been measuring it in the bridge). If you have a term that depends on the head of another project then we'll look up that project each time we evaluate a term, which we typically do hundreds of times each time the metadata branch is updated. So I'd expect this to make evaluation of such configurations much slower. I think @avsm uses these terms in mirage-ci.

samoht: But the caching is already done in the [engine](https://github.com/docker/datakit/blob/cd060d151465e37fab8419674f960893efce1403/ci/src/cI_engine.ml#L84) right?

talex5:
@samoth: (moving down here to avoid having the discussion attached to an outdated diff)

For each project that engine watches, it only scans the github metadata once for each update (not once per term). But when a term refers to a different project (e.g. "build this PR against docker/docker/refs/heads/master"), it looks up the other project ("docker/docker") on each term evaluation. This is what the caching in github_hooks is for.

samoht:
@avsm I have tagged 0.8.0 and created PR on opam-repository:
- [datakit-client](https://github.com/ocaml/opam-repository/pull/7979)
- [datakit-server](https://github.com/ocaml/opam-repository/pull/7980)
- [datakit](https://github.com/ocaml/opam-repository/pull/7982)
- [datakit-ci](https://github.com/ocaml/opam-repository/pull/7981) blocked on https://github.com/cryptosense/multipart-form-data/issues/10

samoht:
Rebased. The client caching code has been merged in #400 so is not needed anymore in the CI code.

samoht:
I have made an other pass on the public API and I decided to rename a few things. Hopefully this won't be too contentious. The main change is changing `Term` into `Job`: I think term is too generic, especially `Term.return`  which is too much dependant on the context in which it is used and hence hard to document/describe. This is now replaced by `Job.success` which is much more explicit IMHO. The other changes are minor tweaks to try to make the API easier to read/grasp. 

The documentation is not finished yet, but this can be improved in another PR. There are already too much changes in that PR, let's merge it :-) (awaiting final review from @avsm and @talex5)

samoht:
I also pushed the doc of that PR online: http://docker.github.io/datakit/Datakit_ci.html

talex5:
Huh? `return` is the standard name for this in all monads. And it's not just "success", it's also "immediate", "unchanging", "ignoring context" and "producing an empty log".

You can't call terms jobs because there is already a thing called a job (calculating the result of a test for GitHub). A job may evaluate many terms. And renaming key concepts in the middle of an unrelated refactoring just makes it even harder to merge :-(

talex5:
(having a `success` that takes a format and evaluates to a successful GitHub status would be fine I think, to go with `fail` and `pending`, but we should keep the generic `return` for terms that aren't evaluating GitHub statuses)

samoht:
Hum ok, I'll revert the `return` and `Term` changes. I am just not fan at all to the too generic `Term` name.

samoht:
The big renaming is reverted and [online doc](http://docker.github.io/datakit/Datakit_ci.html) is updated.

ci/src/datakit_ci.mli
```diff
@@ -0,0 +1,664 @@
+(** Datakit CI
```
talex5: Didn't we decide to use `DataKit` rather than `Datakit` at some point? Also, it's hard to comment on changes to the API now they appear as a new file.

ci/src/datakit_ci.mli
```diff
@@ -0,0 +1,664 @@
...
+    together a collection of job description and a simple description
+    for the {{!web}web} interface, whose access is granted using
+    {{!acl}ACLs}.
+
+    TODO.
+*)
+
+(** {1:core Core} *)
+
+module DK: Datakit_S.CLIENT with type error = Protocol_9p_error.error
```
talex5: All these formatting changes (" : " to ": ") waste a lot of reviewing time. Can we just keep the style the project is already using? Especially for code that already exists - there's no change here except to replace one coding style with a slightly different one.

samoht:
I am merging that patch without a full review. Please ping me if you spot any regression on running systems.

ci/src/datakit_ci.mli
```diff
@@ -0,0 +1,664 @@
...
+  (** [enter_with_pending_reason t msg use fn] is like [use fn], but
+      posts [msg] as the pending reason until [fn] is called (or [use]
+      fails).  This is useful to give a pending reason while getting a
+      mutex or pool resource. *)
+
+  val contents: t -> string
+  (** [contents t] is the current contents of the buffer. *)
+end
+
+module Output: sig
```
talex5: This seems like a bad name. It's the log generated by the term, not the actual output of the term (i.e. the value the term produces).

ci/src/datakit_ci.mli
```diff
@@ -0,0 +1,664 @@
...
+  val use: t -> ?log:Live_log.t -> ?label:string -> job_id ->
+    (unit -> 'a Lwt.t) -> 'a Lwt.t
+  (** [use t job fn] evaluates [fn ()] with one pool resource held.
+      [job] (and [label]) will be displayed as the reason why the
+      resource is in use. If [log] is provided then a message will be
+      logged if we have to wait, and if the log is cancellable then
+      the user will be able to cancel the operation. *)
+end
+
+module Git: sig
```
talex5: Note: this was separate to demonstrate how to write a plugin using only the public API.

PR #389, opened 2016-12-02
github: fix status comparison

samoht:
Thanks to @avsm for spotting this.


bridge/github/datakit_github.ml
```diff
@@ -300,6 +306,7 @@ module Status = struct
       compare_repo;
       compare_commit_hash;
       compare_context;
+      compare_uri;
       Pervasives.compare;
```
avsm: why keep the Pervasives compare here? We know it will always fail

samoht: Updated.

PR #397, opened 2016-12-07
Commit on changes

samoht:
@avsm: this should solve your issue with the bridge pushing empty commits in a loop.


avsm:
awesome, thanks!

PR #399, opened 2016-12-08
datakit: improve the memory representation of paths

samoht:

### DIFF
--- a/src/datakit/main.ml
+++ b/src/datakit/main.ml
@@ -45,11 +45,26 @@ let make_task msg =
   let date = Int64.of_float (Unix.gettimeofday ()) in
   Irmin.Task.create ~date ~owner:"datakit <datakit@docker.com>" msg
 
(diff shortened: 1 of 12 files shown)


src/datakit/main.ml
```diff
@@ -45,11 +45,26 @@ let make_task msg =
...
+  type t = string
+  let equal = equal
+  let compare = compare
+  let hash = hash
+  let to_json = to_json
+  let of_json = of_json
+  let size_of = size_of
+  let write = write
+  let read = read
+  let merge _ = merge []
```
talex5: I'm assuming we're not using this. Maybe raise an exception if it gets called?

samoht: the first parameter is the full path for the file being merged, we indeed don't use that at all in DataKit and I plan to remove that argument in Irmin 1.0 anyway (as in the vast majority of cases, you don't care).

PR #400, opened 2016-12-08
client: add client-side caching

samoht:
Ideally we should cache the subtrees as well, but it is not easy
to expose this with the current API.

### DIFF
--- a/src/datakit-client/datakit_client_9p.ml
+++ b/src/datakit-client/datakit_client_9p.ml
@@ -397,22 +397,87 @@ module Make(P9p : Protocol_9p_client.S) = struct
   end
 
   module Tree = struct
-    type t = { fs : FS.t; path : string list; }
-    let of_id fs id = { fs; path = ["trees"; id] }
-    let read t path = FS.read_node t.fs (t.path /@ path)
-    let stat t path = FS.stat t.fs (t.path /@ path)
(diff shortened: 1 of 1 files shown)


src/datakit-client/datakit_client_9p.ml
```diff
@@ -397,22 +397,87 @@ module Make(P9p : Protocol_9p_client.S) = struct
...
-    let read_link t path = FS.read_link t.fs (t.path /@ path)
+
+    type value = [ `Dir of string list | `File of Cstruct.t | `Link of string ]
+    type 'a cache = 'a Protocol_9p.Error.t Datakit_path.Map.t ref
+
+    type t = {
+      fs   : FS.t;
+      path : string list;
+      reads: value cache;
+      stats: Datakit_S.stat option cache;
```
talex5: What is None vs not in cache?

src/datakit-client/datakit_client_9p.ml
```diff
@@ -397,22 +397,87 @@ module Make(P9p : Protocol_9p_client.S) = struct
...
+    }
+
+    let find_cache c p =
+      try Some (Datakit_path.Map.find p !c) with Not_found -> None
```
talex5: We should probably just get `Datakit_path.Map` to use `Asetmap`.

PR #402, opened 2016-12-09
ci: cope with lost redis connections

talex5:
This isn't quite ideal, but it should allow the CI to recover if the connection to Redis is lost.

See: https://github.com/0xffea/ocaml-redis/issues/44


talex5:
(don't know why CircleCI is failing, but master is failing with the same error so ignoring)

PR #404, opened 2016-12-12
ci: report number of tags, branches and open PRs

talex5:
We recently had a case of a user accidentally pushing a large number of
tags to their repository and wondering why the CI was so busy building
things. It would have been easier to see what the cause was with these
metrics.


talex5:
Use #408 instead.

PR #407, opened 2016-12-12
ci: adapt to protocol-9p.0.8.0

samoht:


talex5:
I think Docker's caching is too clever for you...

PR #408, opened 2016-12-12
More metrics

talex5:
Report number of tags, branches and open PRs

We recently had a case of a user accidentally pushing a large number of
tags to their repository and wondering why the CI was so busy building
things. It would have been easier to see what the cause was with these
metrics.

Report metrics for log messages

It's useful to see e.g. graphs of warning and error log messages.

This metric uses multiple labels (level and src), and includes a fix to make that work.


talex5:
Travis failures seem unrelated (no deps changed in this PR):

The following dependencies couldn't be met:
  - comparelib -> type_conv < 112.02.00

PR #409, opened 2016-12-12
ci: ignore unreadable sessions

talex5:
If we can't parse the saved data for a session, generate a fresh one. This can happen if we change the format (as happened in #384).


samoht:
LGTM

talex5:
Travis failed, but looks unrelated.

```
The following dependencies couldn't be met:
  - comparelib -> type_conv < 112.02.00
```

(this PR doesn't change any dependencies)

samoht:
I believe the TravisCI error is due to https://github.com/ocaml/ocaml-ci-scripts/pull/145 which has been reverted since then.

PR #410, opened 2016-12-13
ci: fix a regression when parsing reference in the webui

samoht:
This was introduced by d1fc7b77

### DIFF
--- a/ci/src/cI_web_templates.ml
+++ b/ci/src/cI_web_templates.ml
@@ -89,7 +89,7 @@ let escape_ref path =
   Uri.pct_encode ~scheme:"http" (String.concat ~sep:"/" path)
 
 let unescape_ref s =
-  Uri.pct_encode ~scheme:"http" s |> (fun s -> String.cuts ~sep:"/" s)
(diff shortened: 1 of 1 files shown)


ci/src/cI_web_templates.ml
```diff
@@ -89,7 +89,7 @@ let escape_ref path =
   Uri.pct_encode ~scheme:"http" (String.concat ~sep:"/" path)
 
 let unescape_ref s =
-  Uri.pct_encode ~scheme:"http" s |> (fun s -> String.cuts ~sep:"/" s)
+  Uri.pct_decode s |> (fun s -> String.cuts ~sep:"/" s)
```
talex5: Shouldn't we split on `/` before decoding?

samoht: You are of course right

samoht: Actually I am not sure to understand. We used to decode and the split in d1fc7b7, I'm just re-doing the same thing here.

talex5:
Tested and works - LGTM.

PR #411, opened 2016-12-13
ci: change the url scheme for PRs and references

samoht:
The new scheme is:

- `/:user/:repo/pr/:id` for PRs
- `/:user/:repo/ref/*` for other references


samoht:
This is incomplete and requires more work.

samoht:
It's ready now. I am now super happy at having the ref encode and ref decode functions in different modules but I don't see where to put them otherwise.

ci/src/cI_web.ml
```diff
@@ -277,8 +280,11 @@ let routes ~logs ~ci ~server ~dashboards =
     ("branch",          fun () -> new branch_list t);
     ("tag",             fun () -> new tag_list t);
     (* Individual targets *)
-    ("pr/:user/:repo/:id",                   fun () -> new pr_page t);
-    ("ref/:user/:repo/:id",                  fun () -> new ref_page t);
+    (":user/:repo/pr/:id",     fun () -> new pr_page t);
+    (":user/:repo/ref/:id",    fun () -> new ref_page None t);
+    (":user/:repo/branch/:id", fun () -> new ref_page (Some "heads") t);
```
talex5: Why do we have two routes to the same page?

ci/src/cI_web.ml
```diff
@@ -161,7 +161,10 @@ class pr_page t = object(self)
         Wm.continue (CI_web_templates.target_page ~csrf_token ~target jobs) rd
 end
 
-class ref_page t = object(self)
+let unescape_ref s =
+  Uri.pct_decode s |> (fun s -> String.cuts ~sep:"/" s)
```
talex5: Agree it would be good to keep this next to `escape_ref`.

ci/src/cI_web.ml
samoht: you mean `ref/$encode(heads/master)` and `branch/master` ? I can filter out the first one, but I think it is more consistent to keep it if client are just interested on raw refs (as some of the refs are neither `branch` or `tags`, e.g. `pull`)

talex5: Yes, I don't see any point in having `branch/foo` when we already have `ref/heads/foo` that does exactly the same thing. It's confusing.

samoht: Because it is a useful distinction for the users? Why do we have separate tabs for tags and branches for instance?

samoht:
Updated: 
* I've removed the `branch` and `tag` route
* I've move the `pct_decode` function close to the `pct_encode` 
* I've modified the way pct-encoding for references is done: `/` are preserved to get nicer-looking urls.

PR #421, opened 2016-12-21
ci: migrate self-ci from docker-compose to docker-cloud

talex5:
The CI is now running in AWS, managed by docker-cloud.


avsm:
I still use docker-compose for the mirage deployment, fwiw

PR #425, opened 2016-12-29
ci: allow non-TLS deployments

talex5:
This is useful if the CI is fronted by another service that provides TLS, to avoid having to renew two sets of certificates.


samoht:
It's also useful for local deployment! :-)

PR #429, opened 2017-01-04
ci: clone missing Git repositories automatically

talex5:
`Git.v ~remote local_dir` will automatically clone `remote` into `local_dir` if `local_dir` does not exist.

This makes it easier to deploy a CI to a new node.

/cc @avsm


samoht:
This also make easier to deploy local CI. Does it work with the `remote` address is a local path?

talex5:
@samoht: yes - in fact the unit tests make use of this.

avsm:
hurrah!

PR #430, opened 2017-01-05
[ci] expose more functions on `Target.v`

avsm:
These are convenience functions that match the equivalents for
the unresolved `Target.t`, used in mirage-ci

### DIFF
--- a/ci/src/cI_target.ml
+++ b/ci/src/cI_target.ml
@@ -87,6 +87,10 @@ let path = function
 
 type v = [ `PR of PR.t | `Ref of Ref.t ]
 
+let pp_v f = function
+  | `PR pr  -> Fmt.pf f "prs/%a" PR.pp pr
+  | `Ref x -> Fmt.pf f "refs/%a" Ref.pp x
+
 let head = function
   | `PR x  -> PR.commit x
   | `Ref x -> Ref.commit x
@@ -96,3 +100,7 @@ let compare_v (x:v) (y:v) = Elt.compare (x :> Elt.t) (y :> Elt.t)
 let path_v = function
   | `PR pr -> path (`PR (PR.id pr))
   | `Ref r -> path (`Ref (Ref.id r))
+
+let repo_v = function
+  | `PR pr -> PR.repo pr
+  | `Ref r -> Ref.repo r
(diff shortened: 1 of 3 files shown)


ci/src/cI_target.ml
```diff
@@ -87,6 +87,10 @@ let path = function
 
 type v = [ `PR of PR.t | `Ref of Ref.t ]
 
+let pp_v f = function
+  | `PR pr  -> Fmt.pf f "pr/%a" PR.pp pr
+  | `Ref x -> Fmt.pf f "refs/%a" Ref.pp x
```
talex5: It seems strange that one case is plural and the other isn't.

Looks like we use `pr` and `ref` in URLs, but `prs/heads/tags` when parsing. Git uses `.git/refs`.

samoht: yes not sure where this is used. I've tried to not use `s` in the datakit/github bridge, but indeed Git is putting `s` everywhere.

avsm: I'm only using this pp_v function for branch name calculation, so the values don't seem significant (but should be consistent). I'll modify it to be plural to match Git then.

avsm:
the CI seems to have failed for reasons other than this patch.

samoht:
minor nitpick: can you change back to use singular for `pr` and `ref` to be more consistent with the rest of the bridge?

talex5:
What causes this?
```
[WARNING] External solver failed with inconsistent return value. Request saved
          to "/home/opam/.opam/log/solver-error-19-1.cudf"
[ERROR] External solver failure, please fix your installation and check
        /home/opam/.opam/config and variable $OPAMEXTERNALSOLVER.
        You may also retry with option --use-internal-solver
```

avsm:
@talex: https://github.com/docker/datakit/pull/433#issuecomment-271265784

avsm:
@samoht I think `pp_v` here should be consistent with `pp` in the same module, which uses `prs` and not `pr`.

talex5:
> CentOS and Fedora are the only containers that use the Docker Cloud-hosted solver.ocaml.io, so you will see that error if the infra is down. The builds can be restarted.

But these are alpine images.

samoht:
@avsm ha yes you are right. I'm fine with the `s` then :-)

avsm:
> But these are alpine images.

I've been migrating the Alpine images to use a builtin solver since Alpine 3.5 -- there was a brief period where they were using the cloud solver too (and pre Alpine 3.5 images still use the cloud solver since there is no backport). More at https://github.com/ocaml/opam/issues/2809#issuecomment-270889218

PR #437, opened 2017-01-11
Updates for Lwt 3 API changes

talex5:
See: https://github.com/ocsigen/lwt/issues/308

### DIFF
--- a/_tags
+++ b/_tags
@@ -1,5 +1,5 @@
 true : bin_annot, safe_string
-not <ci/**>: warn_error(+1..49), warn(A-3-4-41-44)
+not <ci/**>: warn_error(+1..49-3), warn(A-4-41-44)
 true: package(bytes lwt astring logs result cstruct fmt rresult)
 
 <src>: include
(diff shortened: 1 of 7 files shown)


_tags
```diff
@@ -1,5 +1,5 @@
 true : bin_annot, safe_string
-not <ci/**>: warn_error(+1..49), warn(A-3-4-41-44)
+not <ci/**>: warn_error(+1..49-3), warn(A-4-41-44)
```
talex5: @samoht: not sure I quite understand what this does. I think ideally we want:
- Show deprecation warnings when compiling normally, but don't abort.
- Make warnings fatal under CI.

samoht: Your change manages to fix your first bullet, but I don't know how to achieve the second point nicely.

talex5: We could maybe add an ocamlbuild rule (this is what 0install does):

`if ci_build then flag ["compile"; "ocaml"] (S [A"-warn-error"; A"..."]);`

Would be nice to automate it somehow for all projects.

aantron: FWIW Lwt [does something similar](https://github.com/ocsigen/lwt/blob/3ae61f535b341c2d9bf39a911dd1f30313b2f05e/myocamlbuild.ml#L77-L84).

PR #439, opened 2017-01-12
Enable Prometheus reporting from the main database service

talex5:
Run with `--listen-prometheus 9090` to expose metrics at `http://0.0.0.0:9090/metrics`.

Note: there is no encryption and no access control. Front it with nginx or similar if you want to collect metrics remotely.

### DIFF
--- a/pkg/pkg.ml
+++ b/pkg/pkg.ml
@@ -3,13 +3,27 @@
 #require "topkg"
 open Topkg
 
+let includes = function
+  | "prometheus" | "prometheus-app" -> ["prometheus"]
+  | "datakit-ci" -> ["ci"]
+  | "datakit" -> ["src"; "src/datakit"]
+  | "datakit-server" -> ["src"; "src/datakit-server"]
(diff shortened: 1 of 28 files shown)


pkg/pkg.ml
```diff
@@ -3,13 +3,27 @@
...
 open Topkg
 
+let includes = function
+  | "prometheus" | "prometheus-app" -> ["prometheus"]
+  | "datakit-ci" -> ["ci"]
+  | "datakit" -> ["src"; "src/datakit"]
+  | "datakit-server" -> ["src"; "src/datakit-server"]
+  | "datakit-github" -> ["src"; "src/datakit"]
+  | "datakit-client" -> ["src"; "src/datakit-client"]
+  | x -> failwith ("Unknown includes for package: " ^ x)
```
talex5: @samoht: Note: I've changed the build so that each package only sees its own files. Hopefully this should be more reliable, and we can have more than one file called `main.ml`...

PR #440, opened 2017-01-13
ci: Docker plugin

talex5:
This plugin provides the ability to build Docker images, including builds where one built image is used as the base image for another. This is needed to test DataKit itself (needed to make #439 go green!).

Note that @avsm has a more sophisticated Docker plugin at https://github.com/avsm/mirage-ci/ for when the Dockerfile is generated by the CI. This PR's plugin handles the simple case of a project which already has a hand-written Dockerfile.


ci/src/cI_docker.ml
```diff
@@ -0,0 +1,143 @@
+open! Astring
+open Lwt.Infix
+open CI_utils.Infix
+
+module DK = CI_utils.DK
+
+module Image = struct
+  type t = {
+    id : string;
```
avsm: It would be useful to unify this with https://github.com/avsm/mirage-ci/blob/master/src/docker_build.mli#L14 so that we have one type for Docker images within the CI.  I separated out the sha256 (which is queried after the image is built) from the textual tag, and added a human-readable string as well to describe the purpose of the file (this is probably optional).

ci/src/cI_docker.ml
```diff
@@ -0,0 +1,143 @@
...
+    | None -> Fmt.strf "Build %s in %a" t.label CI_git.Commit.pp_short src
+    | Some base -> Fmt.strf "Build %s in %a from %a" t.label CI_git.Commit.pp_short src Image.pp_short base
+
+  let load _t tree _key =
+    DK.Tree.read_file tree CI_cache.Path.value >>*= fun data ->
+    Lwt.return (Image.v (String.trim (Cstruct.to_string data)))
+
+  let branch_safe_char = function
+    | ':' -> '-'
+    | x -> x
```
avsm: Could `branch_safe_char` be exposed elsewhere for other plugins to use? Or is there any reason why the results of `branch` shouldnt always be run through this for all plugins?

ci/src/cI_docker.ml
```diff
@@ -0,0 +1,143 @@
...
+
+  let branch t {Key.src; from} =
+    let from =
+      match from with
+      | None -> ""
+      | Some from -> "-from-" ^ String.map branch_safe_char (Image.id from)
+    in
+    Printf.sprintf "docker-build-%s-of-%s%s" t.label (CI_git.hash src) from
```
avsm: Can be `Fmt.strf` for consistency

ci/src/cI_docker.ml
```diff
@@ -0,0 +1,143 @@
...
+
+  let rewrite_from ~log ?from path =
+    match from with
+    | None -> Lwt.return ()
+    | Some base ->
+      Lwt_io.with_file ~mode:Lwt_io.input path (fun ch -> Lwt_io.read ch) >>= fun contents ->
+      match String.cut ~sep:"\n" contents with
```
avsm: Do we need to filter for `String.is_prefix ~affix:"#"` here as well, to account for comments before the `FROM` ?

ci/src/cI_docker.ml
talex5: Depending on the way the branch is generated, replacing characters automatically could make the branch names non-unique. Possibly we should define some custom escaping scheme (in Irmin) and apply it everywhere.

ci/src/cI_docker.ml
talex5: Yes, I think so.

talex5:
This one does store the sha256 after the build. It uses the somewhat hacky method of building it twice - once normally to get log output and then again with `-q` to get the image ID, relying on the caching to make the second build fast.

talex5:
After discussion with @avsm I'm tempted to merge this now, as the builds are broken without it. The comment thing should be supported at some point, but we might prefer to use a proper Dockerfile parser rather than hacky string matching, and the current code will give a clear error if it gets an unsupported format.

avsm:
lgtm, I dont think this clashes with anything I'm doing. I've created https://github.com/avsm/ocaml-dockerfile/issues/2 to track the Dockerfile parser request in the library.

PR #442, opened 2017-01-14
add Prometheus package to the datakit-ci dependency list

avsm:
Not including this results in a link error against datakit-ci
otherwise.

Signed-off-by: Anil Madhavapeddy <anil@docker.com>


talex5:
Ah, yes. This fix is also in #440, but it makes sense to fix it immediately. Need to update the CI to test for this...

PR #453, opened 2017-01-19
ci: when redirecting to the final logs, use a temporary redirect

talex5:
The problem case was:

- The web UI shows the log as building.
- By the time you click the link, it has finished.
- The CI redirects the browser to the saved results.
- You click rebuild and try to view the new in-progress log.
- The browser remembers the redirect and shows the old saved log instead.


talex5:
Ignoring Travis build failures (one doesn't say what the problem was, but just ends at `opam install ocamlfind`; the other is hanging at `Processing  6/9: [irmin-watcher: ocaml test]`). The fix passed DataKitCI, and that's good enough for me ;-)

PR #454, opened 2017-01-19
ci: improve and simplify metrics collection

talex5:
- Remove support for `metrics_token` (no longer needed since #452).
- Configure self-ci to use the new mechnism to report metrics for the CI and bridge.


talex5:
Closing for now. Most of this PR has already been merged, but we can't remove `metrics_token` until one last user no longer needs it.

PR #462, opened 2017-01-24
Replace use of Travis for CI

talex5:
This PR updates self-ci to run the tests too. This is a bit tricky at the moment as opam likes to install far too many dependencies when testing. It should get faster with opam 2.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,4 @@
-*.native
-*.byte
-_build
-setup.data
-setup.log
+**/*.native
+**/*.byte
+**/_build
+ci/self-ci/data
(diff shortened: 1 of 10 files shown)


samoht:
@talex5 the tests things should be fixed by https://github.com/ocaml/ocaml-ci-scripts/pull/149

talex5:
@samoht: how do I use those scripts here? I'm just doing `opam install -t --deps-only datakit-ci && opam reinstall -v -t datakit-ci` currently.

samoht:
Hum yes in this case, let's wait for opam2. Note: I think @avsm already has some Docker images for opam2, maybe we can just switch to use them.

PR #464, opened 2017-01-26
ci: evaluate terms in parallel

talex5:
This PR allows up to 50 terms to be evaluated at the same time. This may be useful if you have a high-latency link to the database. Testing with SelfCI against a DB in another country, this improved startup time from about 90s to 45s.

/cc @avsm

### DIFF
--- a/ci/src/cI_cache.ml
+++ b/ci/src/cI_cache.ml
@@ -71,22 +71,37 @@ module Make(B : CI_s.BUILDER) = struct
     val find : t -> string -> B.value status option
     val set : t -> string -> B.value status -> unit
     val remove : t -> string -> unit
+    val with_lock : t -> string -> (unit -> 'a Lwt.t) -> 'a Lwt.t
   end = struct
+    type entry = {
+      lock : Lwt_mutex.t;
+      mutable value : B.value status option;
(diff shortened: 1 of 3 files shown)


samoht:
Why 50? You can set unlimited fids when connecting to the server if needed, see https://github.com/docker/datakit/blob/master/bridge/github/main.ml#L134 for instance

talex5:
Unlimited buffering means increased latency. e.g. if you have 1000 9p requests queued up and the user tries to view a log, they'll have to wait for the queue to clear first. Also, we just want to keep the server busy, not overload it. The ideal value will depend a bit on the deployment, but 50 should be plenty I think.

samoht:
My point is that you might need to increase the number of fids that the server will accept for the client (by default it is set to 10 I think) if you want your 50 requests to really be processed in parallel.

samoht:
(my bad, it is set to 100 by default: https://github.com/mirage/ocaml-9p/blob/master/lib/protocol_9p_client.ml#L552)

talex5:
Worth increasing it anyway (I didn't realise the default was still 100). I've also changed the cache to use per-branch locks, which helps a lot.

PR #465, opened 2017-01-30
ci: update to use released version of ocaml-session

talex5:


talex5:
(can't test this at the moment due to the current docker-cloud outage)

PR #476, opened 2017-02-09
Shorter commit messages

samoht:


bridge/github/datakit_github.ml
```diff
@@ -317,19 +319,17 @@ module Status = struct
 
   let pp_opt k ppf v = match v with
     | None   -> ()
-    | Some v -> Fmt.pf ppf " %s=%s" k v
+    | Some v -> k ppf v
```
talex5: This now looks the same as `Fmt.option`.

talex5:
CI tests failed. Also, looks like this needs rebasing.

PR #477, opened 2017-02-09
ci: add user/repo/commit/ID page

talex5:
This means we can make the link for a commit point to the commit page,
which avoids fights if two targets have the same head.

### DIFF
--- a/ci/src/cI_engine.ml
+++ b/ci/src/cI_engine.ml
@@ -87,6 +87,7 @@ type project = {
   canaries : CI_target.Set.t option;
   mutable open_prs : target PR.Index.t;
   mutable refs : target Ref.Index.t;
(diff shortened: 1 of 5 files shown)


talex5:
@samoht : CircleCI keeps failing with 
```
[ERROR] Dependency errors in the upgrade actions. Please update, and report the
        following to the package maintainers if the error persists:
  - remove prometheus-app.~unknown -> remove prometheus-app.~unknown
  - install prometheus-app.~unknown -> install prometheus-app.~unknown
```

I always fix it by clicking "Rebuild without cache". Is there some way to fix it properly?

samoht:
Maybe the cache of the `master` branch needs to be flushed properly. I'll try to click on various buttons...

PR #478, opened 2017-02-09
self-ci: shrink Docker image size

talex5:
Build in a dev container and then copy binary to a fresh container.
Reduces image size from 2.37 GB to 141 MB.


ci/self-ci/Dockerfile.run
```diff
@@ -0,0 +1,6 @@
+FROM alpine:3.5
+RUN apk update && apk add libev docker gmp
```
samoht: maybe add a `--no-cache` option there?

talex5: Ah, didn't know about that - handy!

PR #480, opened 2017-02-09
Split datakit-github into datakit-github and datakit-bridge-github

samoht:
datakit-github provides abstraction API for GitHub and doesn't need a lot of
depencencies. This can be used by the CI and by the local bridge without
requiring to depend on ocaml-github at all.


samoht:
All green!

PR #483, opened 2017-02-09
ci: fix the META file after the datakit-github change

avsm:
see #480


talex5:
Did https://github.com/docker/datakit/pull/482 remove the need for this?

samoht:
@talex5 nope, `datakit-github` actually doesn't depend on `ocaml-github`, it just defines some types and signatures.

samoht:
I will retag 0.9.0 with the fix.

PR #489, opened 2017-02-16
Record history of builds

talex5:
Each time a job's output changes to a new state, it is recorded in a
branch named after the target. There is one directory per job, with an
output file holding JSON of the status and the log tree.

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -8,11 +8,7 @@ RUN opam init -y
 RUN opam install ocaml -y
 RUN opam install depext -y
 
-RUN opam pin add -yn datakit-client.dev https://github.com/docker/datakit.git && \
-    opam pin add -yn github --dev && \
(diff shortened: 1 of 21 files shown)


talex5:
@samoht: that seems like an odd thing to do. You'd still have to parse the JSON to understand which file applied to which build step anyway.

samoht:
Do the steps have names? I am not sure what's the best UI, but that seems a bit weird to me to have a JSON file for targets, when we have files and directories for other branches. Can't every step just be a directory inside the target branch? I am just trying to find a nice way to get target results using the Git CLI (I don't propose to remove the other branches).

talex5:
No, steps don't have names. Also, reading and writing lots of small files is really inefficient (both for Git and for 9p).

talex5:
AppVeyor seems to be failing for an unrelated reason:
```
- File "src/datakit/ivfs_tree.mli", line 14, characters 20-223:
- Error: The signature constrained by `with' has no component named value
```

PR #490, opened 2017-02-21
ci: record an index from commits to builds

talex5:
This allows us to locate archived build results, not just current ones. In particular, this means that old status links from GitHub won't lead to page-not-found errors (at least for builds done after this change).


talex5:
(can't work out how to fix AppVeyor problem, but master was failing with the same problem so ignoring)

PR #492, opened 2017-02-28
Better error for invalid job names

### LINKED ISSUE #371: CI: job names cannot have slashes in them
Building a Term name with a forward slash fails dynamically when it runs through DataKitCI:

```
DataKitCI: internal error, uncaught exception:
           (Invalid_argument "'/' in path step \"2: Revdeps/Compilers\"")
```

talex5:
Before, the CI would stop and log an error. Now, it displays "invalid-name" in the web UI and shows the error message as the job's result.

Fixes #371

### DIFF
--- a/ci/src/cI_config.ml
+++ b/ci/src/cI_config.ml
@@ -5,7 +5,7 @@ type test = string CI_term.t
 
 type project = {
   dashboards : CI_target.Set.t;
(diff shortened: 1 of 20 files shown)


samoht:
@talex5 can you rebase that PR if it is still needed (or close it otherwise?). Thanks!

PR #501, opened 2017-03-23
ci: remove the iframe with the logs

samoht:
This doesn't work at all on small and very large screens.


talex5:
If you rebase on master then the tests should pass (need to pin an older version of opam2).

samoht:
Rebased

PR #505, opened 2017-03-28
Batch up git pushes

talex5:
Before, we ran a single "git push" command each time a branch changed.
If lots of branches were changing, we could spend all our time pushing.

Now, we:

- Wait 2 seconds after a branch becomes dirty in case other branches
  need pushing around the same time.

- Push all dirty branches at once.

I haven't yet measured how much this helps.


samoht:
Thanks! LGTM

PR #508, opened 2017-03-29
ci: include test name in links from GitHub

talex5:
Add the test name to the details URL so that we can highlight the correct test automatically.

Also, ensure aspcud is used during the build to get better dependency resolution.

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -1,7 +1,7 @@
 #FROM ocaml/opam-dev:alpine-3.5_ocaml-4.04.0
 FROM ocaml/opam-dev@sha256:573b5d71e94104590a8282a9ae67b63a6144b15bafdacb9980391c84ca730615
 ENV OPAMERRLOGLEN=0 OPAMYES=1
-RUN sudo apk add tzdata
+RUN sudo apk add tzdata aspcud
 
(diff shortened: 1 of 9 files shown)


datakit-ci.opam
```diff
@@ -9,12 +9,12 @@ dev-repo:     "https://github.com/docker/datakit.git"
 doc:          "https://docker.github.io/datakit/"
 
 build: [
-  "ocaml" "pkg/pkg.ml" "build" "--pinned" "%{pinned}%" "--tests" "false" "-n" name
+  "ocaml" "pkg/pkg.ml" "build" "--dev-pkg" "%{pinned}%" "--tests" "false" "-n" name
```
samoht: are you sure this is needed?

talex5: Not needed, but https://github.com/dbuenzli/topkg/blob/master/CHANGES.md#v090-la-forclaz-vs says `Deprecate --pinned in favor of --dev-pkg in the build command`.

talex5:
CI failed because:
```
Your request can't be satisfied:
  - unavailable for unknown reasons (this may be a bug in opam)
```
This usually means no aspcud. But I think the newer Docker image with aspcud also has a newer opam-dev that can't install test dependencies.

PR #522, opened 2017-04-19
Use nested builds to shrink images

talex5:
When hub supports this, it reduces the image sizes quite a bit:
    
datakit : 1.4 G -> 21 M
github  : 1.7 G -> 22 M


talex5:
Note: tests fail are failing in the CI because the CI tries to install the test dependencies in the final image, which no longer has the build environment. However, the tests are now being run in the main dockerfile (in the build part), so we can remove that from the CI once this is merged.

talex5:
Merging as hub builds have stopped anyway due to the move to the new organisation. Once we have a new org set up, I can get the CI to push there directly.

PR #523, opened 2017-04-19
Go: Separate user config from defaults in the database

djs55:
Previously the database had a single `master` branch containing a mixture of user-specified configuration; runtime internal state and defaults. This meant that in Docker for Mac

- the "schema" upgrade can't tell the difference between a user setting and a default setting, so it resets everything, leading to user suffering and hotfixes. In practice this means we can't use the schema upgrade mechanism any more.
- some keys have an ad-hoc user override mechanism (e.g. the DNS and HTTP proxy keys)
- we don't have a way to change a default for some users (e.g. to enable an experimental feature) while allowing them to opt out by turning it off

These patches use branches to separate the notions of

- user overrides stored in the `master` branch as before
- system defaults stored in the `defaults` branch
- experimental features stored in the `experimental` branch
- the configuration in-effect computed from the previous 3 branches in the `state` branch

On application start we populate the `defaults` branch (and later we will populate the `experimental` branch).

On upgrade we scan the existing `master` branch and identify (by ugly string matching on commit messages) those keys which have not been changed by the user. These keys are deleted so that the keys in the `default` (or `experimental`) branch are used in preference.

When the UI calls `setvmsettings` these overrides are written to the `master` branch. When the UI calls `getvmsettings` the derived values are read from the `state` branch.

`vpnkit` has been modified to read its configuration from the `state` branch. For the moment we write the Moby-relevant keys also to the `master` branch pending a future update to moby which will look at the `state` branch instead.

To see this code in action, download the latest Docker for Mac from the edge channel and take a look at the database directory.

### DIFF
--- a/api/go-datakit/client.go
+++ b/api/go-datakit/client.go
@@ -265,3 +265,55 @@ func (f *FileReader) Read(p []byte) (int, error) {
 	}
 	return n, err
 }
+
+type ioFileReaderWriter struct {
+	f      *File
+	ctx    context.Context
+	offset int64
+}
+
+// NewIOReader creates a standard io.Reader at a given position in the file
+func (f *File) NewIOReader(ctx context.Context, offset int64) io.Reader {
+	return &ioFileReaderWriter{f, ctx, offset}
+}
+
+// NewIOWriter creates a standard io.Writer at a given position in the file
+func (f *File) NewIOWriter(ctx context.Context, offset int64) io.Writer {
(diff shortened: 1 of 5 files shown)


talex5:
Probably needs rebasing to fix CI errors.

talex5:
@djs55 I think you're the only user of this code, so feel free to merge :-)

djs55:
@talex5 I've not got write access so I can't merge :) Could you do it for me?

PR #529, opened 2017-04-22
datakit: update to latest versions of Irmin/Mirage/Cdmliner

samoht:
This patch make Irmin 1.1 works with DataKit. As Irmin 1.1 only
works with cmdliner >= 1.0 and MirageOS >= 3.0, also update
these dependencies.

The main change is the removal of the body Ivfs_tree to use the new
Irmin API. The new Irmin Tree API is pretty similar (on purpose) but
it seems that we are loosing caching of file hashes. I plan to re-add
that in a later commit, after checking it is really needed.

Ideally datakit-client will also move to something closer to the Irmin API
but it is not a blocker for API compatibility.

### DIFF
--- a/src/datakit/ivfs_blob.ml
+++ b/src/datakit/ivfs_blob.ml
@@ -3,7 +3,7 @@ open Result
 type t = Cstruct.t list ref (* (reversed) *)
 
 let pp_buf ppf buf = Fmt.string ppf (Cstruct.to_string buf)
-let pp ppf t = Fmt.pf ppf "%a" Fmt.(list ~sep:(unit "") pp_buf) !t
+let pp ppf t = Fmt.pf ppf "%a" Fmt.(list ~sep:nop pp_buf) !t
 
 (* FIXME: very expensive! *)
 let compare x y =
--- a/src/datakit/ivfs_blob.ml
+++ b/src/datakit/ivfs_blob.ml
@@ -2,8 +2,8 @@ open Result
 
 type t = Cstruct.t list ref (* (reversed) *)
 
-let pp_buf ppf buf = Fmt.pf ppf "%S" (Cstruct.to_string buf)
-let pp ppf t = Fmt.pf ppf "%a" Fmt.(Dump.list pp_buf) !t
+let pp_buf ppf buf = Fmt.string ppf (Cstruct.to_string buf)
+let pp ppf t = Fmt.pf ppf "%a" Fmt.(list ~sep:(unit "") pp_buf) !t
 
 (* FIXME: very expensive! *)
 let compare x y =
@@ -27,6 +27,8 @@ let to_ro_cstruct t =
   let cs = Cstruct.concat (List.rev !t) in
   t := [cs];
   cs
+
+let t = Irmin.Type.(like cstruct) of_ro_cstruct to_ro_cstruct
 
 let to_string t =
   Cstruct.to_string (to_ro_cstruct t)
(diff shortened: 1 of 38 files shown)


samoht:
Note: this needs the dev version of hvsock, + https://github.com/mirage/irmin/pull/432 and https://github.com/mirage/irmin/pull/431 to be merged first (and Irmin 1.1 to be released)

talex5:
CI is unhappy because the `depext` line also adds some (conflicting) pins.

samoht:
I've updated the PR with @talex5 comments, but I haven't fixed the CI yet. Working on that now.

src/datakit/ivfs_blob.ml
```diff
@@ -2,8 +2,8 @@ open Result
 
 type t = Cstruct.t list ref (* (reversed) *)
 
-let pp_buf ppf buf = Fmt.pf ppf "%S" (Cstruct.to_string buf)
-let pp ppf t = Fmt.pf ppf "%a" Fmt.(Dump.list pp_buf) !t
+let pp_buf ppf buf = Fmt.string ppf (Cstruct.to_string buf)
+let pp ppf t = Fmt.pf ppf "%a" Fmt.(list ~sep:(unit "") pp_buf) !t
```
avsm: I think the `unit ""` can be replaced by `nop` from Fmt

src/datakit/main.ml
```diff
@@ -41,65 +41,47 @@ let error fmt = Printf.ksprintf (fun s ->
...
-  let hash = hash
-  let to_json = to_json
-  let of_json = of_json
-  let size_of = size_of
-  let write = write
-  let read = read
-  let merge _ = merge []
-  module Path = Ivfs_tree.Path
-end
+  Irmin.Info.v ~date ~author:"datakit <datakit@docker.com>" msg
```
avsm: could we remove the `datakit@docker.com` email entirely? This is now a `moby/datakit` repo, but could the default commit could also just be `datakit` without an email address if that's valid in Git?

src/datakit_io.ml
```diff
@@ -116,26 +117,93 @@ module FS = struct
...
         aux (Filename.dirname dir) >>= fun () ->
-        Log.debug (fun f -> f "mkdir %s" dir);
+        Log.debug (fun l -> l "mkdir %s" dir);
         protect (Lwt_unix.mkdir dir) 0o755;
       ) in
     Lwt_pool.use mkdir_pool (fun () -> aux dirname)
 
+  let file_exists f =
+    Lwt.catch (fun () -> Lwt_unix.file_exists f) (function
+        (* See https://github.com/ocsigen/lwt/issues/316 *)
```
avsm: This issue is now fixed in Lwt>=2.7.1 -- update constraint and remove this workaround?

src/datakit/main.ml
samoht: Email is mandatory, so Irmin will add the default `irmin@openmirage.org` email to it. But yes `@docker.com` should go away.

src/datakit_io.ml
samoht: True. Note that this is a copy/paste from Irmin-unix to avoid depending on nocrypto and hence libgmp, so it needs to be fixed upstream too.

samoht:
The `ci/datakit/github` is legit, it's a regression in Irmin 1.0. Should be fixed by https://github.com/mirage/irmin/pull/438

samoht:
Yay, all green! 

Need a rebase of protocol-9p (that @talex5 is working on) before it can be merged.

talex5:
Try pinning my `ping-mirage-3` ocaml-9p branch instead.

samoht:
All green, merging!

PR #530, opened 2017-04-25
self-ci: all moby members to rebuild

talex5:
/cc @samoht 

(note: can't test this myself as I'm not a member)


talex5:
(CircleCI failed, but it's not testing this anyway so must be unrelated)

PR #532, opened 2017-04-25
Use Jbuilder

samoht:
This is a work in progress but so far it enables to speed the compilation of all packages from 50s down to 8s.

The release bits are still managed by `topkg`.

The missing bits:
- [x] CI tests
- [ ] watermarking in dev mode
- [ ] doc generation
- [ ] (maybe?) nicer alcotest output for tests (with colors!)
- [ ] ~self-ci, example, skeleton~


samoht:
It's all green!

I am fine with loosing temporary the watermarking in dev mode as @diml seems interested to add that to `jbuilder` directly. 

For the docs, I am trying to find a replacement for publishing docs to `gh-pages`, probably using odoc/odig directly. @dbuenzli: what would be your recommendation to replace the current doc delegate (I guess this is releated to https://github.com/dbuenzli/topkg/issues/24)?

samoht:
I am also fine to temporary break the doc generation, as this should be fixed later on by `odoc` rules directly in `jbuilder`.

@talex5 are you ok if we merge this in its current state?

dbuenzli:
> @dbuenzli: what would be your recommendation to replace the current doc delegate (I guess this is releated to dbuenzli/topkg#24)?

What do you mean ? There's no such thing as a doc delegate.

samoht:
Hum ok. So there is no way to have an alternate build command to build the doc?

dbuenzli:
Read `topkg help doc`.

samoht:
Yup I read this bit:WARNING. The way this command works is at the moment very ad-hoc and
       ocamlbuild specific. It will change in the future.

samoht:
Merging, I will try to workaround the doc generation before the next release.

PR #543, opened 2017-04-28
fix release script

samoht:

### DIFF
--- a/ci/self-ci/selfCI.ml
+++ b/ci/self-ci/selfCI.ml
@@ -19,7 +19,7 @@ module Dockerfile = struct
   let datakit     = v ~timeout:(30. *. minute) "Dockerfile"
 end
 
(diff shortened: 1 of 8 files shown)


ci/self-ci/selfCI.ml
```diff
@@ -80,7 +80,7 @@ module Tests = struct
         "local-git",   check_builds images#local_git;
         "libraries",   Term.wait_for_all [
           "client",      check_builds images#client;
-          "client-4.02", check_builds images#client_4_02;
+          "client-4.04", check_builds images#client_4_04;
```
talex5: Note that `client` already tests with 4.04.

samoht: Fixed in https://github.com/moby/datakit/pull/546

PR #551, opened 2017-05-09
9p split

samoht:
client/server: split the libraires between core API and 9p transport

There is now:
  - datakit-client: signature for client API + Path library
  - datakit-server: implementation of the VFS on top of Irmin
  - datakit-client-9p: implementation of the API using 9p as transport
  - datakit-server-9p: expose the Irmin VFS as the Datakit API; server-side
    implementation of the API using 9p as transport

The tests are split as well, so all the client/server tests can be re-used
with a different transport mechanism.


samoht:
That commit contains also some consistency cleanups for `Datakit_S.CLIENT`. @talex5 let me know if you prefer these in a separate PR.

avsm:
would be good to cut a release before this gets merged, to help with constraints.

samoht:
@avsm: ok I'll cut a new release

samoht:
@avsm: see https://github.com/ocaml/opam-repository/pull/9122

samoht:
The CI tests are now becoming flaky, need to understand what happens before merging. Seems that an exception is escaping somehow:

```
[...]
2017-05-09 13:56.38 INF [datakit-ci] recalc_thread got abort request
2017-05-09 13:56.38 INF [application] Disconnecting 9p
Fatal error: exception End_of_file
Raised at file "string.ml", line 118, characters 19-34
Called from file "src0/sexp.ml", line 90, characters 13-47
```

samoht:
I still don't really understand which async thread is sometimes throwing the `End_of_file` exception when a server is turned down. I believe this is an artifact of the tests (which start/stop servers in a loop) so should be safe to ignore as done by 019237d.

I'll merge this once all is green.

talex5:
If we can't debug where unknown exceptions are coming from, that in itself sounds like a bug. I might make a patch to Lwt to add an optional label argument to `Lwt.async`, but we can wrap it ourselves for now. Other options:

- Run with backtraces turned on `OCAMLRUNPARAM=b`.
- Run with tracing enabled (see https://github.com/mirage/mirage-profile#recording-traces-manually).

talex5:
Following the EOF thread back mirage-trace-viewer, it looks like it's coming from a 9p client thread:

eof-9p-client

GitHub won't let me upload the trace itself, but I'll send you a copy.

samoht:
@talex5: thanks! The error seems indeed to come from https://github.com/mirage/ocaml-9p/blob/master/lib/protocol_9p_server.ml#L183. When that line is protected, the `End_of_file` error goes away.

talex5:
I think this is the exception we're seeing:

https://github.com/mirage/ocaml-cstruct/blob/master/lwt/lwt_cstruct.ml#L31

I think we need to change `Flow_lwt_unix.write` to return an error code in this case.

samoht:
Thanks to @talex5's detective work, a proper fix has been pushed upstream: https://github.com/mirage/ocaml-9p/pull/126

PR #564, opened 2017-05-17
datakit: use Irmin.Merge.idempotent instead of re-implementing our own

samoht:


samoht:
Seems that CircleCI's VM is broken, merging anyway as the rest is green.

PR #571, opened 2017-05-18
Fix warnings in tests

samoht:


samoht:
Note: this caught the fact that we were not running the full test suite for the bridge. Fixed now (and the tests still pass)

PR #573, opened 2017-05-18
client 9p: build the mount helper in the right library

samoht:


samoht:
CircleCI fix is in #574

PR #577, opened 2017-05-20
Allow the GH bridge to run without a server (and use a local Git directly)

samoht:
For instance, you could do:

```
$ datakit-bridge-github -r samoht/test -d git:///tmp/foo
Starting datakit-bridge-github %%VERSION%% (*:rw)...
Connecting to git:///tmp/foo [github-metadata].
```

And then:

```
$ tree /tmp/foo
.
└── samoht
    └── test
        ├── pr
        │   └── 32
        │       ├── base
        │       ├── head
        │       ├── state
        │       └── title
        └── ref
            ├── heads
            │   ├── adassad
            │   │   └── head
            │   ├── bar
            │   │   └── head
            │   ├── master
            │   │   └── head
            │   ├── samoht-patch-1
            │   │   └── head
            │   ├── test
            │   │   └── head
            │   ├── test2
            │   │   └── head
            │   ├── toto
            │   │   └── head
            │   └── yyy
            │       └── head
            └── tags
                └── titi
                    └── head

16 directories, 13 files
```

You can then edit the Git repo, and when you commit changes they are reflected directly to GitHub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,4 @@
-FROM ocaml/opam-dev@sha256:64dc0522876ebbd27b186e3ba955ae5ab864ace580add5b0d1abb8715bdf1bbe
-RUN git -C /home/opam/opam-repository fetch origin && \
-    git -C /home/opam/opam-repository reset origin/master --hard && \
-    opam update -u
+FROM ocaml/opam-dev@sha256:1dd4440b3e5f182f705cb5a74f9d4e860c2842b45ed72c199de89a894d13f522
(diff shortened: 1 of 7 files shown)


samoht:
The failure is due to https://github.com/moby/datakit/blob/master/ci/src/cI_docker.ml#L125

From @talex5 on slack:
> Ah, maybe we need a third case:
> - No `FROM` override => pull
> - Locally-build from => don't pull
> - Published image => pull

samoht:
I am not super fond on how to specify the type of the endpoint, but let's change that in a subsequent PR if that's needed.

PR #578, opened 2017-05-26
add List func in Snapshot to get files in directory

ebriney:
- return filelist as []string
 - add test

Signed-off-by: ebriney <emmanuel.briney@docker.com>


talex5:
(CI failures are unrelated, so merging)

PR #596, opened 2017-06-08
github: rename PR.owner into PR.user

samoht:


samoht:
Closing as this is actually not a good idea. `PR.user` is already there and means "user or org" where the repo which the PR has been made live.

PR #601, opened 2017-06-13
ci: document the key and context types better

talex5:
/cc @samoht


samoht:
Thanks!

Issue #597, opened 2017-06-09
Invalid use of `${ROOT}`


diml:
While analyzing packages using jbuilder I found a few use of `${ROOT}` in datakit to point to the root of the project. This doesn't work in general, as the expansion of `${ROOT}` depends on the workspace configuration. This means that you can't build datakit when it is a subdirectory in a bigger workspace.

To help with this, I added `${SCOPE_ROOT}` in the development version, which always point to the first parent directory containing `<package>.opam` files (see [here](http://jbuilder.readthedocs.io/en/latest/project-layout-specification.html#scopes) for more details).

talex5:
(looks like this isn't in 1.0+beta10, so we'll need to wait for the next release)

samoht:
Fixed by hardcoding the path instead of using `SCOPE_ROOT`, but fixed anyway.

PR #604, opened 2017-06-23
build: support jbuilder workspaces

avsm:
remove use of ${ROOT} and use relative path referencing, which
in turn allows datakit to be built as part of a bigger jbuilder
workspace.

This can be tested by cloning this repository into a subdir
and running `jbuilder build` in the top-level directory.

Signed-off-by: Anil Madhavapeddy <anil@docker.com>


samoht:
Thanks for the patch, this fixes #597. The CI is failing for unrelated reasons, so merging.

PR #605, opened 2017-07-01
Go: fix the handling of defaults over upgrade

djs55:
Originally we had a single database branch and interpreted
- a key being present: a user's setting which must be preserved
- a missing key: a "default" value which should be filled in

Since we split the configuration into multiple branches we should always write our default values to the "defaults" branch. The user config goes in the "master" branch, any experimental features are in the "experimental" branch and the final results go in the "state" branch.

This PR
- updates the "defaults" branch over upgrades
- uncomments and fixes one of the unit tests
- adds a new unit test to cover the new desired behaviour

### DIFF
--- a/api/go-datakit/config.go
+++ b/api/go-datakit/config.go
@@ -140,43 +140,29 @@ func (r *Record) Upgrade(ctx context.Context, schemaVersion int) error {
 	return r.Wait(ctx)
 }
 
-// fillInDefault writes the default value to the store if no Value
-// is already present. This ensures that the system state is always
-// in sync with the database, and we don't have to also know what
-// default values are also baked into the application.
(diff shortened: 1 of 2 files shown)


djs55:
I think the failures are all OCaml related e.g.
```
The command '/bin/sh -c opam depext -i asl win-eventlog alcotest mtime mirage-flow hvsock     git irmin irmin-unix lwt protocol-9p-unix tyxml redis multipart-form-data     pbkdf tls prometheus-app github git session irmin irmin-unix cmdliner     webmachine alcotest' returned a non-zero code: 1
[2017-07-01 20:36.54] Failed: Command "docker" "build" "--pull" "-f" "Dockerfile.ci" "." exited with status 1
```

PR #611, opened 2017-09-28
Ignore trailing newlines when doing 'test and set' on reference files

### LINKED ISSUE #610: "git commit" seems to break the transaction interface
If I create an empty database:
```
mkdir database
cd database
git init .
```
then start datakit:
```
com.docker.db --git <path>/database  -url file:///var/tmp/foo
```
and then run the Go unit tests:
```
cd go-datakit
go test
```
-- everything passes.

If I make a commit to the database directory with `git`:
```
cd database
echo foo > bar
git add bar
git commit -m test
```
then (whether I restart the server or not) running the tests hangs committing a transaction:
```
$ go test
2017/09/27 22:17:38 Testing the client interface
2017/09/27 22:17:38 Dialling unix /var/tmp/foo
2017/09/27 22:17:38 Remove [branch master transactions foo rw a b c]
2017/09/27 22:17:38 Testing the configuration interface
2017/09/27 22:17:38 Dialling unix /var/tmp/foo
```
The test which triggers this seems to be:
```
$ go test -run TestTransaction
2017/09/27 22:18:00 Testing the transaction interface
2017/09/27 22:18:00 Dialling unix /var/tmp/foo
```
If I add some printfs:
```
$ git diff .
diff --git a/api/go-datakit/transaction_test.go b/api/go-datakit/transaction_test.go
index f59aaef..b49912f 100644
--- a/api/go-datakit/transaction_test.go
+++ b/api/go-datakit/transaction_test.go
@@ -24,8 +24,10 @@ func TestTransaction(t *testing.T) {
        if err != nil {
                t.Fatalf("Transaction.Write failed: %v", err)
        }
+       log.Println("Commit")
        err = trans.Commit(ctx, "Test transaction")
        if err != nil {
                t.Fatalf("Transaction.Commit failed: %v", err)
        }
+       log.Println("Done")
 }
```
then I get
```
$ go test -run TestTransaction
2017/09/27 22:19:09 Testing the transaction interface
2017/09/27 22:19:09 Dialling unix /var/tmp/foo
2017/09/27 22:19:09 Commit
```
which means that the `Commit` operation is hanging.

samoht:
This is because `git commit` might add these. Fix #610

/cc @djs55

### DIFF
--- a/src/datakit-io/datakit_io.ml
+++ b/src/datakit-io/datakit_io.ml
@@ -397,12 +397,24 @@ module FS = struct
           )
       )
 
+  let trim s =
+    let len = Cstruct.len s in
+    if len >= 2
+    && Cstruct.get_char s (len - 1) = '\n'
+    && Cstruct.get_char s (len - 2) = '\r'
+    then
+      Cstruct.sub s 0 (len - 2)
+    else if len >= 1 && Cstruct.get_char s (len - 1) = '\n' then
+      Cstruct.sub s 0 (len - 1)
+    else
+      s
+
   let test_and_set_file ?temp_dir ~lock file ~test ~set =
     Lock.with_lock (Some lock) (fun () ->
         read_file file >>= fun v ->
         let equal = match test, v with
           | None  , None   -> true
-          | Some x, Some y -> Cstruct.equal x y
+          | Some x, Some y -> Cstruct.equal (trim x) (trim y)
           | _ -> false
         in
         if not equal then Lwt.return false


samoht:
The proper fix should probably be to modify https://github.com/mirage/ocaml-git/blob/master/src/git/fs.ml#L630-L638, but this should be fine too.

djs55:
@samoht I've got some CI fixes on #612 for CircleCI and appveyor -- feel free to grab any you don't already have!

djs55:
I think you might need
- xcode.8.3: https://github.com/moby/datakit/pull/612/commits/30c52ade0a146664ba7c9aa1e9b18c8abdb95366
- remove one of the references to version.ml: https://github.com/moby/datakit/pull/612/commits/ba1f3203a52d0316e347df6ecf81c8477b0001b6
- fix opam-depext no such file or directory: https://github.com/moby/datakit/pull/612/commits/b6b6aa18d0ba5cf583c2f65893a896f80b3a60bc
- fix "cannot find topfind" https://github.com/moby/datakit/pull/612/commits/52fa58cf50fdcbcbce2b0d043e2c0313de857a64

(the `opam config exec --` ones are a bit weird. There was a bug in OCaml's standard library recently to do with environment variables wasn't there?)

I think these should fix CircleCI and appveyor

samoht:
@djs55 thanks, I have rebased on top on your branch

Dockerfile.ci
```diff
@@ -1,8 +1,7 @@
-FROM ocaml/opam-dev@sha256:1dd4440b3e5f182f705cb5a74f9d4e860c2842b45ed72c199de89a894d13f522
-#FROM ocaml/opam-dev:alpine-3.5_ocaml-4.04.0
+FROM ocaml/opam:alpine-3.5_ocaml-4.04.0
```
talex5: Why the move back to opam 1?

samoht: The CI changes are now worked on #612 where I reverted that change. The plan is to first merge #612 and then rebase this PR on master with just the "newline" commit on it.

samoht:
Rebased on top of master.

PR #612, opened 2017-09-28
attempt to fix CI

djs55:
Using Xcode 8.3 in `circle.yml` should hopefully fix the error:

> carthage: A full installation of Xcode.app 8.3 is required to compile this software.
> Installing just the Command Line Tools is not sufficient.
> Xcode can be installed from the App Store.
> Error: An unsatisfied requirement failed this build.

Removing one of the references to the `version` module in `bridge/github` should fix the error:

> Multiple rules generated for _build/default/bridge/github/version.o

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,12 @@
-FROM ocaml/opam-dev@sha256:1dd4440b3e5f182f705cb5a74f9d4e860c2842b45ed72c199de89a894d13f522
+FROM ocaml/opam-dev@sha256:f04c7fbe8fe103b77186f02a5b6c43f9a52e31af26b1211adf86698cfb5bb91f
 #FROM ocaml/opam-dev:alpine-3.5_ocaml-4.04.0
 
 RUN git -C /home/opam/opam-repository fetch origin && \
-    git -C /home/opam/opam-repository reset 482bd5f55 --hard && \
+    git -C /home/opam/opam-repository reset 3bcfe6c8d --hard && \
     opam update -u
 
 ENV OPAMERRLOGLEN=0 OPAMYES=1
-RUN sudo apk add tzdata aspcud gmp-dev perl
+RUN sudo apk update && sudo apk add tzdata aspcud gmp-dev perl
 
 RUN opam depext -ui lwt inotify alcotest conf-libev asl win-eventlog \
     irmin-watcher mtime mirage-flow conduit hvsock prometheus-app git irmin \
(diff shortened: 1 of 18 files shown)


samoht:
I have pushed more CI fixes directly to your branch. Let's make this green :-)

Makefile
```diff
@@ -23,8 +23,7 @@ depends:
 	  datakit-ci datakit-bridge-github datakit-bridge-local-git -y
 
 datakit:
-	$(BUILD) -n datakit
-	$(RUNTEST) test/datakit
+	$(BUILD) -p datakit
```
talex5: Why did this go? Are we testing it elsewhere?

samoht: We are not testing this. And `jbuilder -n datakit` just says that `-n` is not a valid option.

samoht:
I have fixed the Dockerfiles and updated everything to use the latest version. Not sure why appveyor and CircleCI are complaining, need to have a closer look.

samoht:
I've bumped appveyor and CircleCI to 4.05, hopefully this will fix the last issues.

samoht:
The appveyor built is blocked on:
```
datakit -> hvsock >= 0.8.1 -> duration -> alcotest < 0.8.0
```
which should be solved once https://github.com/ocaml/opam-repository/pull/10770 is updated to the windows repository.

Dockerfile
```diff
@@ -1,12 +1,12 @@
...
 #FROM ocaml/opam-dev:alpine-3.5_ocaml-4.04.0
 
 RUN git -C /home/opam/opam-repository fetch origin && \
-    git -C /home/opam/opam-repository reset 482bd5f55 --hard && \
+    git -C /home/opam/opam-repository reset 3bcfe6c8d --hard && \
     opam update -u
 
 ENV OPAMERRLOGLEN=0 OPAMYES=1
-RUN sudo apk add tzdata aspcud gmp-dev perl
+RUN sudo apk update && sudo apk add tzdata aspcud gmp-dev perl
```
thaJeztah: `apk add --no-cache` should work as well?

samoht:
@talex5 would it be possible to remove the CI tests (as they are now tested with the main `Dockerfile.ci`) and to remove the check for the client library on 4.03 (or to use an image with opam2 instead of opam1)?

talex5:
@samoth: I've pushed a fix to https://github.com/talex5/datakit/tree/fix-ci and deployed it.

samoht:
@talex5 Thanks! Now all seems to be green apart the windows builds as we are waiting for https://github.com/fdopen/opam-repository-mingw to be updated.

samoht:
Finally all green! Merging.

PR #637, opened 2018-09-23
Add support for the latest Atd 2.0 JSON adapters

avsm:
These have a new interface for handling unknown tags, which changes
the polymorphic variant interface in the github bridge in datakit.

This also fixes the build by adding a constraint on Irmin which
has changed how it hashes, and explicitly depending on named-pipe.lwt

This PR depends on `opam pin github --dev`


samoht:
LGTM. CircleCI is failing as it has to be ported to opam2.

PR #640, opened 2018-12-14
circleci: update to v2.0

djs55:
CircleCI 1.0 is not supported any more.


.circleci/config.yml
```diff
@@ -0,0 +1,35 @@
...
+      MACOSX_DEPLOYMENT_TARGET: "10.10"
+      TERM: vt100
+    steps:
+      - checkout
+      - run: brew uninstall python # remove files in /usr/local/bin
+      - run: brew install wget ocaml opam pkg-config dylibbundler
+      - run: opam init -n https://github.com/ocaml/opam-repository.git
+      - run: opam switch create 4.05
+      - run: opam switch 4.05
+      - run: opam pin add datakit-server.dev . -n
```
avsm: I thnk you can do `opam pin add -n .` here to get all of them in one go

.circleci/config.yml
```diff
@@ -0,0 +1,35 @@
...
+      - run: opam pin add datakit-server-9p.dev . -n
+      - run: opam pin add datakit.dev . -n
+      - run: opam update && opam upgrade --fixup && opam upgrade
+      - run: opam install depext && opam config exec -- opam depext osx-fsevents datakit
+      - run: opam list
+      - run: opam config exec -- opam reinstall datakit-server datakit -v
```
avsm: you could also do `opam install .` at this point to install them all i think

.circleci/config.yml
```diff
@@ -0,0 +1,35 @@
...
+      - run: mkdir -p _build/src/datakit
+      - run: opam config exec -- make bundle COMMIT
```
avsm:
```suggestion
      - run: opam exec -- make bundle COMMIT
```

djs55:
This fixes the CircleCI build although there is still work to do on the other CIs.

avsm:
Looks like the datakit CI is using and old opam2 snapshot?

```
Step 1/16 : FROM ocaml/opam-dev:alpine-3.5_ocaml-4.05.0
alpine-3.5_ocaml-4.05.0: Pulling from ocaml/opam-dev
```

That should be replaced by `ocaml/opam2:alpine`

PR #641, opened 2019-02-28
datakit/ci: adapt to crunch 3.0.0

hannesm:
crunch 3.0.0 removes the `size`, which is included here only to silence warning 32. crunch 3.0.0 is PR'ed to opam-repository https://github.com/ocaml/opam-repository/pull/13549


avsm:
Looks like the CI dockerfile is using an older snapshot and needs a `git pull`. @djs55 do you have access to this?

PR #644, opened 2019-03-11
support the webmachine 0.6.0+ interface

avsm:
This however conflicts with irmin-http, which requires webmachine<0.6.0.

So this PR is a draft until Irmin 2.0.0 is released which has a newer irmin-http.
Unless we can cut a minor release of irmin-http from the 1.x branch @samoht?

PR #646, opened 2019-04-14
autoformat all code using ocamlformat 0.9

avsm:
This uses the `conventional` profile to minimise surprise, and brings the style inline to the autoformatting conventions being adopted by the OCaml community.

Since datakit is essentially stable now, I also took the liberty of bumping the release version of this to 1.0.0 so that it can be parked.


samoht:
LGTM

avsm:
I'll hold off on the 1.0 release until we have the latest webmachine (post irmin 2.0) from #644

